PR #2, opened 2016-10-30
Add support for RocksDB and LMDB

bauerj:


kyuupichan:
Thanks Johann!

I'd prefer to have the 4 storage files in a single storage.py.   Do you have anything against that?

Two other things:

- can you make the super() calls just super().whatever(); super itself shouldn't need any args I think
- can you add yourself to AUTHORS with a comment about providing a backend DB abstraction, or however you want to refer to it.

If the above is good please resubmit a pull req & I'll merge it.

bauerj:
Sure 👍 I updated the PR according to your requests.

kyuupichan:
Thanks, merged.

PR #3, opened 2016-11-05
Edited NOTES to include HOST env

SuBPaR42:


kyuupichan:
Thanks, but unfortunately I only saw it after releasing 0.06 with something similar.

PR #23, opened 2016-11-13
Only count clients that actually interact with the server 

bauerj:
Clients that did not send a single command are probably not electrum clients and maybe connected by accident.

### DIFF
--- a/server/protocol.py
+++ b/server/protocol.py
@@ -156,7 +156,7 @@ class ServerManager(LoggedClass):
         return self.irc.peers
 
     def session_count(self):
-        return len(self.sessions)
+        return len([s for s in self.sessions if s.send_count > 0])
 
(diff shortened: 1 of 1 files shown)


kyuupichan:
What about the sessions list?  Do we want to filter them there?

bauerj:
I'm not sure about that. On one hand, it might confuse people if the count doesn't add up but on the other hand it could be useful to diagnose problems (especially if logs are disabled).

My primary concern was about monitoring the session count. I do have a lot of clients that just connect and do nothing (`Sent 0 bytes in 0 messages`).

Maybe adding the `send_count` to the session list would be a good compromise?

kyuupichan:
I've done this slightly differently in develop branch.  Let me know if it could be better for you

PR #24, opened 2016-11-13
Add optional option to anonymise logs (remove IP addresses)

### LINKED ISSUE #20: Option to anonymize logs?
Is it possible to add an option to anonymize logs so that ip addresses are changed to x.x.x.x or something?

bauerj:
This will replace IP addresses reported when connecting or disconnecting with the string `XX.XX.XX.XX`.

Session.peername gets a new parameter called `for_log` that should be True whenever the return value is intended to be stored in a log file.

This doesn't affect the RPC call `sessions`, IP addresses will still be shown there.

Closes #20


kyuupichan:
Thanks - just one thing, can you make the env var and prog vars ANON_LOGS / anon_logs, it avoids the s/z issue.

bauerj:
Done 👍

kyuupichan:
Thanks.

PR #25, opened 2016-11-13
Add unit tests for the methods in util

bauerj:
Add unit tests for the `util` module.


kyuupichan:
Fantastic, thanks.  I can think of a few more; I'll add those

PR #34, opened 2016-11-20
Provides sample systemd env file, and organizes init system files

cluelessperson:


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bauerj: As much as I like IntelliJ, this should probably not be in the repository.

samples/init system/systemd/electrumx.conf
```diff
@@ -0,0 +1,102 @@
...
+    USERNAME = electrumx
+    ELECTRUMX = /usr/local/bin/electrumx_server.py
+
+    #Bitcoin Node Credentials
+    DAEMON_URL = http://username:password@hostname:port/
+        #Alternatively you may specify the individual variables
+        #DAEMON_USERNAME =
+        #DAEMON_PASSWORD =
+        #DAEMON_HOST =
+        #DAEMON_PORT = #optional, will default to coin appropriate default
```
bauerj: This is already obsolete :no_mouth:

samples/init system/systemd/electrumx.conf
```diff
@@ -0,0 +1,102 @@
...
+    #REPORT_TCP_PORT = $TCP_PORT
+    #REPORT_SSL_PORT = $SSL_PORT
+
+
+#
+#OPTIONAL VISIBILITY
+#
+    #BANNER_FILE = /path/to/banner
+    #DONATION_ADDRESS =
+    #IRC = freenode
```
bauerj: IRC=yes would also work 😉

bauerj: To use this, you will also need to install to `irc` package.

kyuupichan:
Just 2 things:

- can you drop the .gitignore change; it seems to be something not relevant to most users
- can you drop the "init system" directory?  I don't want spaces in paths, but I don't think we need that level at all.  Just samples/daemontools and samples/systemd would be good

kyuupichan:
Dup of #36 which went in

PR #35, opened 2016-11-20
Add DEFAULT_RPC_PORT for DASH

thelazier:


kyuupichan:
Thanks, I've merged into my development branch; this will be in the next release.

kyuupichan:
Hello,can you please add a PEERS for dashcoin and maybe testnet?  We're going to be transitioning away from IRC.  It should be a list of known good servers that have stuck around for a while.

thelazier:
Done #117

PR #38, opened 2016-11-22
Add IRC_PREFIX,IRC_CHANNEL,IRC_SERVER,IRC_PORT

thelazier:
This should allow setting up electrum server for other Coin to find peers on other places.
For example, electrum server for DASH should join #electrum-dash with prefix 'D' for Mainnet and 'd' for Testnet.

### DIFF
--- a/docs/ENV-NOTES
+++ b/docs/ENV-NOTES
@@ -52,8 +52,12 @@ MAX_SESSION_SUBS - maximum number of address subscriptions permitted to a
 If you want IRC connectivity to advertise your node:
 
 IRC             - set to anything non-empty
+IRC_PREFIX      - The default is 'E'.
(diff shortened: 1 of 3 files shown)


kyuupichan:
Hi, thanks for looking at this.

There should not be any need for new env vars - COIN and NET should be enough.  All the customisation is a property of the coin.

So let's put IRC_PREFIX, IRC_CHANNEL, IRC_SERVER and IRC_PORT as class variables in lib/coins.  Also please but the underscore in the prefix so the nick is IRC_PREFIX + nick, where for bitcoin the IRC_PREFIX is E_.

Does this seem reasonable?

kyuupichan:
Also if PEER_REGEXP is moving from class var to instance var let's make it lower case again; upper case is for class vars.  Thanks.

thelazier:
replaced with #41

PR #39, opened 2016-11-22
Variables replaced with DAEMON_URL

cluelessperson:


kyuupichan:
I'll fix the old env vars.  I like the README as-is.

PR #45, opened 2016-11-25
fs_block_hashes need correct header_hash from coin

thelazier:
So, I try to replace header_hashes with header_prevhash and header_hash. Then I use header_hash to replace double_sha256 in fs_block_hashes. 
Please let me know if I do something I shouldn't.


kyuupichan:
Actually one is used for the error message, so keep that in a local var but get rid of the other.  Thanks.

kyuupichan:
Perfect, thanks a lot.

PR #51, opened 2016-11-28
Corrected Litecoin's IRC_PREFIX

santzi:
There was typo on Litecoin's IRC_PREFIX.


kyuupichan:
I fixed it when I merged it yesterday

kyuupichan:
Hello,can you please add a PEERS for litecoin and maybe testnet?  We're going to be transitioning away from IRC.  It should be a list of known good servers that have stuck around for a while.

PR #53, opened 2016-12-01
Add IRC option to publish Tor address

valesi:
### IRC
- If a `REPORT_*_PORT` is `0`, then don't publish the protocol for public use

### Tor
- Add IRC nick/client for Tor hidden service with `REPORT_HOST_TOR`
- Publish TCP and SSL ports that ElectrumX listens on, and ignore respective `REPORT_*_PORT` setting


valesi:
Added `REPORT_TCP_PORT_TOR` and `REPORT_SSL_PORT_TOR` which default to their respective `REPORT_*_PORT` setting. If `REPORT_*_PORT` is `0`, then use the listening port.

Setting a `REPORT_*_PORT_TOR` to `0` doesn't publish the protocol for the Tor service at all.

kyuupichan:
Thanks very much; will try this out now.  If it looks good it'll be in 0.7.18

PR #57, opened 2016-12-04
Serve batch responses directly, without buffering

bauerj:
Currently responses to a batch request are buffered in a list and then flushed out to client when the last request handler is finished.

This changes the behaviour to immediately send the response to the client, constructing the array on the wire.

This might be in violation to **6)** of the [JSONRPC specification](http://www.jsonrpc.org/specification), specifically

> The Server should respond with an Array containing the corresponding Response objects, **after all of the batch Request objects have been processed.** 

However, as the client can't tell the difference anyway, it speeds up response times and reduces memory usage, that's probably not important.

### DIFF
--- a/electrumx_rpc.py
+++ b/electrumx_rpc.py
@@ -58,7 +58,7 @@ def main():
     args = parser.parse_args()
 
     if args.port is None:
-        args.port = int(environ.get('ELECTRUMX_RPC_PORT', 8000))
+        args.port = int(environ.get('RPC_PORT', 8000))
 
     loop = asyncio.get_event_loop()
     coro = loop.create_connection(RPCClient, 'localhost', args.port)
--- a/lib/jsonrpc.py
+++ b/lib/jsonrpc.py
@@ -216,7 +216,7 @@ class JSONRPC(asyncio.Protocol, LoggedClass):
         if self.error_count >= 10:
             self.transport.close()
 
-    def send_json(self, payload):
+    def send_json(self, payload, delimiter='\n'):
         '''Send a JSON payload.'''
         # Confirmed this happens, sometimes a lot
         if self.transport.is_closing():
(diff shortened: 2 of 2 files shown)


electrumx_rpc.py
```diff
@@ -58,7 +58,7 @@ def main():
     args = parser.parse_args()
 
     if args.port is None:
-        args.port = int(environ.get('ELECTRUMX_RPC_PORT', 8000))
+        args.port = int(environ.get('RPC_PORT', 8000))
```
bauerj: This is of course unrelated to the change, but in `env` `RPC_PORT` is used instead of `ELECTRUMX_RPC_PORT`.

kyuupichan:
Thanks for this.  I see a couple of issues though which make me prefer a different approach:

- I think you've lost the large request handling, which is important for batches.  I still want to reject them once they pass the limit
- send_count seems to be skipped to.

Mind if I try something a little different?  and thanks for pointing out the RPC_PORT bug

bauerj:
> send_count seems to be skipped to.

Yes, I forgot that. That should probably be added in the loop in `JSONRPC.batch_payload`

> I think you've lost the large request handling, which is important for batches. I still want to reject them once they pass the limit

I think that's still being handled in `send_json`.

> Mind if I try something a little different?

Not at all, go ahead.

kyuupichan:
By large request handling I meant that, it seems the limit is being imposed on the individual sends from the batch, rather than the batch as a whole, unless I'm misreading it.

bauerj:
Yes, that's correct. Why should individual requests be handled differently from batched requests?

kyuupichan:
Hmm.  Let me think about it.

kyuupichan:
I've reworked the JSON layer in 0.8.5.  Please take a look.  If it works as intended it should supersede this.
For batches it converts each part to JSON text as it processes the batch.  This saves memory as JSON text is more compact than the Python data structures they represent.  Bandwidth is also accounted for as each is processed.  If the response being built is too large this is detected during processing and processing is stopped early.  Otherwise the result is then concatenated and sent.

Let me know your thoughts.

kyuupichan:
This code is obsolete now. Related ongoing changes are in the works.

PR #64, opened 2016-12-08
Tweak systemd unit file settings

bauerj:
- Set the open file limit to 8192, which should hopefully be enough
- Set the timeout for process termination to 30 minutes. Systemd will send another SIGTERM after 30 minutes and then a SIGKILL after 60 minutes. That should be plenty of time for electrumx to cleanly finish whatever it's doing


kyuupichan:
Thanks

PR #96, opened 2017-01-08
add digibyte and argentum

protonn:


kyuupichan:
Hi, thanks.  The GENESIS_HASH is now a string, not a binary string.  Can you try vs 0.10.2?

protonn:
fixed, sorry about that.

PR #113, opened 2017-01-30
remove error variable from normal replies in JSONRPCv1 protocol

erasmospunk:
Even if no error occurred, the server adds an "error":null variable to normal replies.

This change was introduced on commit 06c8eda161a2ae6795bacea92e26977a9c37bd79

### DIFF
--- a/lib/jsonrpc.py
+++ b/lib/jsonrpc.py
@@ -61,7 +61,7 @@ class JSONRPCv1(JSONRPC):
     @classmethod
     def response_payload(cls, result, id_):
         '''JSON v1 response payload.  error is present and None.'''
-        return {'id': id_, 'result': result, 'error': None}
+        return {'id': id_, 'result': result}
 
     @classmethod
     def error_payload(cls, message, code, id_):


kyuupichan:
Sorry, that is not the JSON RPC 1.0 spec as I read it.  My comment even makes it explicit.  Your changes are to JSON RPC 1.0.

What you want is JSON RPC 2.0, which is what the library already does for JSON 2.0.

If this is causing an issue with client software please give me details; I'm interested.

erasmospunk:
@kyuupichan this breaks older Coinomi clients that also use the electrum protocol. The reason is a bug incorrectly incorrectly parses `"error":null` as `"error":"null"` and the client assumes that an error occurred.

This was the behavior of the vanilla electrum server (no error variable if no error) and in electrumx until the 06c8eda161a2ae6795bacea92e26977a9c37bd79 commit.

The JSONRPCv1 indeed requires `"error":null`. Is there a way to force the backwards compatible behavior?

kyuupichan:
Hmm, that's unfortunate.  Incidentally I see nothing in the 06c8eda commit that would cause this - are you sure it's the right commit?

Is there any way to identify coinomi clients, or do they pretend to be Electrum clients?
Are newer versions of coinomi affected?
How many older version installations are there still?

I could force JSON RPC v2 for all incoming connections; that probably is OK with Electrum - would that work for both old and new coinomi?  If that works I cannot promise to keep doing it forever.

erasmospunk:
@kyuupichan sorry my mistake, copied the wrong hash. This is the commit a7462a6ea3997ef0404df9212397b571af8b523e.

Practically all versions except the newest version. There are probably some thousands but most auto-update.

So I noticed that before that commit, it was using JSON RPC v2 by default and it was working fine.

Would you like a pull request that adds a new env variable to force a particular RPC version? I think this is the cleanest solution.

kyuupichan:
Well, that means your problem becomes a server operator problem; each will be different.   I'll try and get a release out in the next 15 hrs that fixes it somehow.  Hopefully in a month or two it won't matter and I can fall back to detecting the peer's protocol version.

kyuupichan:
Can you let me know if 0.10.15 which I just released fixes this please.

kyuupichan:
@erasmospunk I will retain this until April but cannot promise beyond that.

erasmospunk:
@kyuupichan It works, thanks for the suggestion. If you like, you can remove the JSONRPCv2 forcing and I will just use a patch.

kyuupichan:
No problem.  It's in 0.10.15, I'll keep it until April.

Unfortunately (or fortunately in this case) 0.10.11-0.10.15 have a bug where they occasionally send JSON RPC 2.0 batch reponses that kill the electrum client, so we're asking those server owners to upgrade to 0.10.16.  That should eliminate most of your issues.  But please ensure your recent software doesn't have the issue as I intend to remove it in the future - it's most robust when client protocol is auto-detected and not forced.  Thanks.

PR #117, opened 2017-02-02
Add PEERS for DASH

thelazier:


kyuupichan:
If you set PEER_DEFAULT_PORTS for testnet (see the bitcoin testnet example) you can omit the explicit setting in the peers list

kyuupichan:
Thanks

PR #118, opened 2017-02-02
Coin fixes

erasmospunk:
- Fixed a crash when a coin didn't define the TX_COUNT_HEIGHT and IRC related variables
- Formatting and some fixes to coins.py

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -43,6 +43,7 @@ class Coin(object):
     # Peer discovery
     PEER_DEFAULT_PORTS = {'t': '50001', 's': '50002'}
     PEERS = []
+    TX_COUNT_HEIGHT = 0
 
     @classmethod
     def lookup_coin_class(cls, name, net):
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -32,16 +32,16 @@ class CoinError(Exception):
 class Coin(object):
     '''Base class of coin hierarchy.'''
 
-    REORG_LIMIT=200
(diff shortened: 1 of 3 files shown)


erasmospunk:
Ops! Fixing...

erasmospunk:
Hmm... seems that the test coverage decreased bit

kyuupichan:
the coins.py changes are good.  Not so keen on the rest.  If the variable isn't there, how about a default in the base class, or raising an exception in env.py?

kyuupichan:
What causes the crash if IRC is not set?  I missed that.

kyuupichan:
Let me think about this a while

kyuupichan:
I can't do 2 of 3.  Can you leave out the IRC; it seems wrong.  The other 2 are good

kyuupichan:
I put 2/3 in 0.10.17; you didn't get back on the IRC part which seems wrong to me.

erasmospunk:
@kyuupichan sorry missed the notifications.

Regarding the IRC, it is crashing [here](https://github.com/kyuupichan/electrumx/blob/master/server/irc.py#L34) for coins that didn't define IRC_CHANNEL (and IRC_PREFIX). The idea was that if IRC is not defined, then the IRC class should be created but I made a mistake in the `main_loop` function, the check should have been `if self.env.irc` as it was before.

kyuupichan:
I see, thanks.  I'd rather we require all these to be set than fudge around them not being set.

So I'll do that.

PR #120, opened 2017-02-05
update hostnames

shsmith:


kyuupichan:
Thanks; I assume those will remain stable for a while

shsmith:
Yes, those host names will be stable.
I also fixed electrum.hsmiths.com to report in the proper case. DNS isn't case-sensitive, but it looked funny with mine being the only one that was capitalized in the seed peers.

PR #121, opened 2017-02-08
recommendations from pycodestyle (pep8 style)

nmarley:
Left a few out where they seemed to make sense, but this is most of the errors caught by the pycodestyle tool.


kyuupichan:
Thanks; I've merged this into my develop branch locally; it will be in 0.11.0 which is coming out soon.

PR #122, opened 2017-02-10
Adding FairCoin

thokon00:
This PR adds FairCoin2 to the list of available coins. Note that FairCoin2 has not been launched yet. But soon :)

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -21,7 +21,7 @@ import sys
 
 from lib.hash import Base58, hash160, ripemd160, double_sha256, hash_to_str
 from lib.script import ScriptPubKey
-from lib.tx import Deserializer, DeserializerSegWit
+from lib.tx import Deserializer, DeserializerSegWit, DeserializerFairCoin
 import lib.util as util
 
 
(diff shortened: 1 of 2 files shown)


thokon00:
Because FairCoin2 has payload data AFTER the transaction data that would trigger this assertion here: https://github.com/kyuupichan/electrumx/blob/master/lib/tx.py#L92

FairCoin2 implements proof-of-cooperation and thus the header differs from 'mining' coins. e.g. We have no bits field instead there is another payload hash. Therefore, header size: 108 (see here: https://github.com/faircoin/faircoin2/blob/master/src/primitives/block.h#L33-L38)

erasmospunk:
@thokon00 do you think you could add a test with a faircoin block deserialization?

kyuupichan:
I see.  So my hack to accomodate the bitcoin genesis coinbase issue is also hurting you hence your own hack for height 0?
I don't have a fundamental issue with your changes but want to think about it a bit more.  My focus has been on the peerdisc branch; sorry for ignoring other issues in the meantime.  Overriding to avoid an assertion seems lame but I'm not sure I have a better idea either.

thokon00:
Thanks for the info! In the meantime I'm going to prepare a test as requested.

lib/coins.py
```diff
@@ -545,3 +545,68 @@ class DigiByteTestnet(DigiByte):
...
+    def header_offset(cls, height):
+        '''Given a header height return its offset in the headers file.
+        If header sizes change at some point, this is the only code
+        that needs updating.'''
+        return height * 108
+
+    @classmethod
+    def header_len(cls, height):
+        '''Given a header height return its length.'''
+        return 108
```
erasmospunk: Please remove this method as the `Coin` superclass calculates it from the `header_offset()`.

lib/tx.py
```diff
@@ -217,3 +217,11 @@ def read_tx(self):
...
         return TxSegWit(version, marker, flag, inputs,
                         outputs, witness, locktime), double_sha256(orig_ser)
+
+
+class DeserializerFairCoin(Deserializer):
+    def read_block(self):
+        '''Returns a list of (deserialized_tx, tx_hash) pairs.'''
+        read_tx = self.read_tx
+        txs = [read_tx() for n in range(self._read_varint())]
+        return txs
```
erasmospunk: Is this deserializer necessary?

The code seems the same as the `Deserializer::read_block` with the `assert self.cursor == len(self.binary)` check. Did you get a crash on that assert?

lib/coins.py
thokon00: True, I'll remove that.

lib/tx.py
thokon00: Yes, it crashed because of the assertion.
A Bitcion block has this structure:
```
header (80 bytes)
list of transactions
```
A FairCoin block has this format:
```
header (108 bytes)
list of transactions
additional proof-of-cooperation payload
```

And this additional payload triggers the assertion. That's why I was forced to create a Deserializer.

kyuupichan: Hi!

Let's just remove the assertion from the base class, then you don't need anything special for your coin in tx.py.  I'll merge the PR if you make that change.

kyuupichan: Oh, and please add a comment where the assertion used to be that some coins have excess data beyond the end of the transactions.  Thanks.

thokon00:
Ready for merge now.

kyuupichan:
Thanks

PR #131, opened 2017-02-18
better handling of bad peers

shsmith:
refrain from adding remote peers from bad peer
log which remote peers are added

### DIFF
--- a/server/peers.py
+++ b/server/peers.py
@@ -77,8 +77,8 @@ class PeerSession(JSONSession):
         proto_ver = (version.PROTOCOL_MIN, version.PROTOCOL_MAX)
         self.send_request(self.on_version, 'server.version',
                           [version.VERSION, proto_ver])
-        self.send_request(self.on_peers_subscribe, 'server.peers.subscribe')
         self.send_request(self.on_features, 'server.features')
+        self.send_request(self.on_peers_subscribe, 'server.peers.subscribe')
 
     def connection_lost(self, exc):
         '''Handle disconnection.'''
@@ -108,6 +108,10 @@ class PeerSession(JSONSession):
                      for real_name in real_names]
         except Exception:
             self.log_error('bad server.peers.subscribe response')
+            return False
+
+        if self.failed:
+            self.log_warning('refrain from calling add_peers for bad peer')
             return False
 
         self.peer_mgr.add_peers(peers)
@@ -143,7 +147,7 @@ class PeerSession(JSONSession):
                 self.peer.update_features(features)
         # For legacy peers not implementing features, check their height
         # as a proxy to determining they're on our network
-        if not verified:
+        if not verified and not self.failed:
             self.send_request(self.on_headers, 'blockchain.headers.subscribe')
         self.close_if_done()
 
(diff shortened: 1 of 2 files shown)


server/peers.py
```diff
@@ -77,8 +77,8 @@ def connection_made(self, transport):
         proto_ver = (version.PROTOCOL_MIN, version.PROTOCOL_MAX)
         self.send_request(self.on_version, 'server.version',
                           [version.VERSION, proto_ver])
-        self.send_request(self.on_peers_subscribe, 'server.peers.subscribe')
         self.send_request(self.on_features, 'server.features')
+        self.send_request(self.on_peers_subscribe, 'server.peers.subscribe')
```
kyuupichan: This doesn't achieve anything - there is no pause between these function calls

server/session.py
```diff
@@ -269,9 +269,13 @@ def server_features(self):
...
-        tor_proxy = self.controller.peer_mgr.tor_proxy
-        peer_info = self.peer_info()
-        return peer_info and peer_info[0] == tor_proxy.ip_addr
+        try:
+            tor_proxy = self.controller.peer_mgr.tor_proxy
+            peer_info = self.peer_info()
+            return peer_info and peer_info[0] == tor_proxy.ip_addr
+        except Exception as e:
+            self.log_error('is_tor: {}'.format(e))
+            return False
```
kyuupichan: I'd like to understand and fix this bit properly; none of it should throw.  I think the cause is a missing transport=None in jsonrpc.py, but not sure

server/peers.py
```diff
@@ -143,7 +147,7 @@ def on_features(self, features, error):
                 self.peer.update_features(features)
         # For legacy peers not implementing features, check their height
         # as a proxy to determining they're on our network
-        if not verified:
+        if not verified and not self.failed:
             self.send_request(self.on_headers, 'blockchain.headers.subscribe')
```
kyuupichan: OK; this line is good; I'll put that in my next update.   I'm not sure what the rest is trying to fix or what you saw; can you elaborate?

server/peers.py
shsmith: What I was concerned about was the order that the requests are popped and processed. I moved the on_peers_subscribe to be last because it needs to know if the on_features result was failure. Before I moved it, on_peers_subscribe would fire before "incorrect genesis hash" was detected.

server/session.py
shsmith: I agree. In fact, I made that a different commit with the intention of not including it in the PR. But my git skills are lacking and my PR ended up with both commits. My bad. I have more examples, if you need them. It only happens rarely and not at all since I made this change.

kyuupichan:
I've incorporated most of these ideas into 0.11.4, but slightly differently.  In particular we only subscribe to peers once the peer is verified.  I think this is the effect you wanted.

server/peers.py
kyuupichan: Understood.  The problem is that there is no guaranteed order of replies (even if that's what it looks like most of the time) and I don't want that to be an implicit assumption for the code to work properly.  The sends are guaranteed to go out in the order written but the remote end is free to reply in any order.
So in my push I did it so that we wait until verification from prior replies and then send the request.

PR #139, opened 2017-02-23
fix tor coin peers always scheduled for check

### LINKED ISSUE #138: PeerManager coin onion peers are always selected for verifying availability
While running with some debugging output for #135 I noticed that coin onion peers are always selected to retry connecting/verify availability. If `tor_proxy.port` is not `None` then `last_tor_retry_time` is never updated.

fr3aker:
fixes #138


kyuupichan:
Thanks

PR #140, opened 2017-02-24
Dynamic header support

erasmospunk:
Block headers can have a dynamic size that is being indexed on a
new meta file "headers_offsets".

The offsets are 64bit in order to accommodate coins with big
headers that will accumulate GBs of header data after some years.
We don't store the offset of the genesis header as it's always 0,
this means that the offset for height 1 is stored at position 0,
for height 2 at 8 and so on.

This commit adds support for AuxPow coins like Namecoin, Dogecoin
and Equihash coins like Zcash


erasmospunk:
Small fix

erasmospunk:
Small refactor

erasmospunk:
Some clean ups

erasmospunk:
Fixed edge case

kyuupichan:
John - I'll be leaving these changes for post 1.0.  They are best reviewed if you can break them up into perhaps 3 or 4 parts.  Let's just do 1 part at a time so you don't have to keep updating later ones if earlier ones change.

erasmospunk:
@kyuupichan just pushed the changes in smaller commits.

Please notify me when you are ready to merge so I can rebase to the latest master if needed.

erasmospunk:
Rebased to the latest master

erasmospunk:
Removed commit #141 from this pull request

erasmospunk:
@kyuupichan changed `raw[0:verlen]` to `raw[:verlen]`. Thanks!

kyuupichan:
John, I've merged your first commit and repushed to master.

erasmospunk:
Rebased

kyuupichan:
Parts 2 and 3 look mostly good.  I want to review those more slowly though, and I'd like to change a couple of things.
In a future PR please add yourself to AUTHORS given the amount you've now submitted.

kyuupichan:
The coins.py push seems to have broken multisig addresses; getting lots of invalid address errors for good addresses beginning with 3

erasmospunk:
Good catch.

erasmospunk:
Rebased to the latest master

erasmospunk:
Rebased to the latest master

lib/tx.py
```diff
@@ -237,3 +264,67 @@ def read_tx(self):
...
+        self.cursor = start
+        return self._read_nbytes(header_end)
+
+
+class TxZcash(namedtuple("Tx", "version inputs outputs locktime")):
+    '''Class representing a transaction.'''
+
+    @cachedproperty
+    def is_coinbase(self):
+        return self.inputs[0].is_coinbase if len(self.inputs) > 0 else False
```
erasmospunk: @kyuupichan this transaction class is the same as `Tx` with the difference that it checks if any inputs are present (the "JoinSplit" transactions have 0 inputs). Should I leave this class or move the check to `Tx`?

kyuupichan:
I'll try and take a look at this stuff this weekend.  Regarding your Tx question, not looked in detail but would probably prefer the tx be a separate class like you have it, perhaps derived from the existing one.

kyuupichan:
John, sorry this is just a bit too much of a large change all at once.  I really appreciate the effort you've put into this and the tests, but I need it broken up a bit.  A few other things:

- Can we do the API changes first?  Then maybe one or more changes, and finally the new coin definitions and tests.
- I know the definition of block will change with the named tuple, but I'd like to keep calling it a block.  So let's just stick with block and blocks rather than b, parsed_blocks, full_block etc.
- I'd like to avoid the Deserializer construction you've done.  I had to stare at it a while (and I learnt something new!).  I don't want tx.py objects having a COIN member for now.  The old classes won't need one; the methods of the new deserializer classes can be passed "cls" to a coin parameter if necessary when invoked from coins.py.  But looking at your changes it seems all they really want is a basic_header_size to be passed.
- perhaps coin.DESERIALIZER should be lower case and just be the function in tx.py for bitcoin and similar, and be a function in the coin class for aux_pow coins
- Do we actually really need a read_header API in tx.py?  Can we keep it as-is with read_tx and read_block (perhaps to be renamed read_txs)?  I'd like to keep tx.py to be just about transactions; I think the knowledge of the block format and header size better belongs in the coin class.
- in db.py I'd like to call fs_update_header_offsets unconditionally, and return doing nothing if STATIC_BLOCK_HEADERS, rather than testing call sites and asserting inside the function.

I realize this might be a bit of work.  If you prefer I can have a go at doing the initial API changes, but I'm not sure I'll get everything.  Then you can follow up with the bulk of the change.  Or you can do the lot.  But if you could break it up into at least 3 pieces it would help.  One at a time is fine.

I've created a dynamic-headers branch on github.  Can you work against that?  Let me know how you'd like to proceed.

erasmospunk:
@kyuupichan thanks for the quality code review.

> Can we do the API changes first? Then maybe one or more changes, and finally the new coin definitions and tests.

Good point. Regarding the coin definitions, what do you think about a folder that contains a .py file per coin? This should keep clean the coins.py

Regarding the other points I will take a look in the next days.

kyuupichan:
Let's keep it all in coins.py for now.  Later something might be more appropriate,indeed.  One advantage to having it all in coins.py is selecting the coin class is easy programatically (the lookup_ function); it might be more awkward if in separate files.

erasmospunk:
@kyuupichan one thing that I noticed, when parsing the block it is useful to have the `Deserializer::cursor` property, especially for parsing a block footer if it exists.

erasmospunk:
Although technically we don't need the footer...

erasmospunk:
Refactored code but haven't tried it yet, please review but don't merge.

One thing that I would like to refactor is to expose the `Deserializer` private methods like `_read_varint` and `_read_nbytes` that are essential to parse the dynamic headers.

Another difference is that we don't read the block footers anymore.

lib/coins.py
```diff
@@ -622,22 +632,14 @@ class FairCoin(Coin):
...
         block and its height.'''
 
-        if height == 0:
-            return []
-
-        deserializer = cls.deserializer()
-        return deserializer(block[cls.header_len(height):]).read_block()
+        if height > 0:
+            return cls.block_full(block, height)
+        else:
```
kyuupichan: This looks like an infinite loop.  I'll try and fix it here.

server/block_processor.py
```diff
@@ -566,14 +567,14 @@ def backup_blocks(self, blocks):
         coin = self.coin
         for block in blocks:
             # Check and update self.tip
-            header = coin.block_header(block, self.height)
+            header, txs = coin.block_full(block, self.height)
             header_hash = coin.header_hash(header)
```
kyuupichan: I'm merging your first commit of the 4 and following up with a couple of tweaks as mentioned above.  If you could rebase that would be great.  But before doing so let me make a couple of comments on the 2nd commit too.

kyuupichan: Oh and that code was good enough to survive a forced testnet 3 block reorg via RPC

server/db.py
```diff
@@ -192,6 +196,35 @@ def clean_db(self):
...
+        self.headers_offsets_file.write(height_start * 8, b''.join(offsets))
+
+    def dynamic_header_offset(self, height):
+        assert not self.coin.STATIC_BLOCK_HEADERS
+        # The offsets are 64 bits in order to accommodate coins with big
+        # headers that will accumulate GBs of header data after some years
+        # We don't store the offset of the genesis header as it's always 0
+        # so this means that the offset for height 1 is stored at
+        # position 0, for height 2 at 8 and so on.
+        if height > 0:
```
kyuupichan: I'd rather waste 8 bytes on disk and not have the extra comment and if statement.  Any reason not to store height 0 as 0 too?  How is it not getting written?  I might be missing something.

And can you add yourself to AUTHORS?  This 2nd commit looks good otherwise.

lib/coins.py
```diff
@@ -293,6 +293,43 @@ def electrum_header(cls, header, height):
...
+            merkle_size = block._read_varint()
+            block.cursor += 32 * merkle_size # Chain merkle branch
+            block.cursor += 4 # Chain index
+            block.cursor += 80 # Parent block header
+            header_end = block.cursor
+        else:
+            header_end = cls.BASIC_HEADER_SIZE
+        block.cursor = start
+        return block._read_nbytes(header_end)
+
```
kyuupichan: Do you think the bulk of the block_header method here is best put in an aux-pow specific deserializer?  I think that's what you did before and I think I prefer it now it's clearer what's happening.  Sorry about that.  Similarly for Zcash below.

So from here just call AuxPowDeserializer.block_header(cls.BASIC_HEADER_SIZE, cls.VERSION_AUX_POW) or something like that.  Those 2 variables belong in the coin class as they are now.  With that I think I'm OK with everything once you rebase.  I'll merge it.  Thanks.

kyuupichan: One more thing, please put in a commit message the github text "Closes #xxx" for the 2 outstanding issues related to this aux pow support.  I assume it does resolve them.

server/db.py
erasmospunk: Yes, block 0 offset is a special case to "bootstrap" the first offset when the `headers_offsets_file` is initially empty, otherwise `dynamic_header_offset` will not work. 

Check how the `fs_update_header_offsets` calculates the offset of block N based on the header size of block N-1.

One pythonic "fix" could be:

```
try:
    pos = (height - 1) * 8
    offset, = unpack('<Q', self.headers_offsets_file.read(pos, 8))
    return offset
except struct.error as e:
    if height == 0:
        return 0
    else:
        raise Exception('Header offset not found: {}'.format(height))
```

Another alternative is to check and write the block 0 offset when creating the `headers_offsets_file`.

kyuupichan: Yes, do you mind writing the initial 0?  There is a similar issue with tx_count, and I don't like my current solution, it caused too many if statements.

erasmospunk:
Updated with the changes you asked. Also I made some tests with Dogecoin, Zcash and it seems to work without an issue like before.

kyuupichan:
Thanks John.  I'll merge this as it seems you're happy with non-BTC stuff.  I'll run this on my server for a while, and if it's good I'll merge it into master.

PR #146, opened 2017-03-13
contribute notes for Raspberry Pi and Ubuntu 16.04 deployment 

shsmith:


contrib/python3.6/python-3.6.sh
```diff
@@ -0,0 +1,11 @@
+Installation of Python 3.6
```
bauerj: This should probably be a comment.

contrib/raspberrypi3/run_electrumx.sh
```diff
@@ -0,0 +1,38 @@
+###############
+# run_electrumx
```
bauerj: Why is this necessary? Isn't systemd the default for Raspbian Jessie?

shsmith: It is not necessary. It is a tip to allow for easier testing and deployment. I'm ok if you would rather delete it.

contrib/python3.6/python-3.6.sh
shsmith: Ok with me.

contrib/raspberrypi3/run_electrumx.sh
bauerj: No, it's fine for me if you think it's useful. I was just confused because it looked like an init-replacement.

Issue #145, opened 2017-03-13
setup.py installation?


atweiden:
I put together a quick [PKGBUILD for electrumx](https://aur.archlinux.org/packages/electrumx). It works by installing the server as a module with `python setup.py install`. The setup script is likely the cause of a startup error. I would assume that the libs would go in `/usr/lib/python3.6/site-packages/electrumx`, but they're being installed to parent directory `/usr/lib/python3.6/site-packages`:

```
$ pacman -Qql electrumx
/etc/
/etc/electrumx/
/etc/electrumx/electrumx.conf
/usr/
/usr/bin/
/usr/bin/electrumx-rpc
/usr/bin/electrumx-server
/usr/lib/
/usr/lib/python3.6/
/usr/lib/python3.6/site-packages/
/usr/lib/python3.6/site-packages/electrumx-1.0-py3.6.egg-info/
/usr/lib/python3.6/site-packages/electrumx-1.0-py3.6.egg-info/PKG-INFO
/usr/lib/python3.6/site-packages/electrumx-1.0-py3.6.egg-info/SOURCES.txt
/usr/lib/python3.6/site-packages/electrumx-1.0-py3.6.egg-info/dependency_links.txt
/usr/lib/python3.6/site-packages/electrumx-1.0-py3.6.egg-info/requires.txt
/usr/lib/python3.6/site-packages/electrumx-1.0-py3.6.egg-info/top_level.txt
/usr/lib/python3.6/site-packages/lib/
/usr/lib/python3.6/site-packages/lib/__init__.py
/usr/lib/python3.6/site-packages/lib/__pycache__/
/usr/lib/python3.6/site-packages/lib/__pycache__/__init__.cpython-36.opt-1.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/__init__.cpython-36.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/coins.cpython-36.opt-1.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/coins.cpython-36.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/enum.cpython-36.opt-1.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/enum.cpython-36.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/hash.cpython-36.opt-1.pyc
/usr/lib/python3.6/site-packages/lib/__pycache__/hash.cpython-36.pyc
[...]

PR #147, opened 2017-03-13
Minimize default systemd config file

bauerj:
The current config file has a few issues:

 - It suggests that a syntax like `COIN = Bitcoin # comment` would do something else than setting `COIN` to `Bitcoin # comment`
 - It doesn't contain all settings
 - It contains some settings that are not used (e.g. `ELECTRUMX`)

To minimize maintainment overhead, I suggest to keep the default config file small and link to the appropriate documentation that should  be up to date.

Closes: #145


kyuupichan:
Agree this is a good idea.  Thanks!

PR #149, opened 2017-03-17
Fix Litecoin parameters

pooler:
- Changed xpub/xprv prefixes to the official values (see e.g. the latest version of [Litecoin Core](https://github.com/litecoin-project/litecoin/tree/v0.13.2.1))
- Added testnet4 parameters
- Added peer lists


erasmospunk:
The values are correct however the Litecoin's testnet4 is not yet released (current version is 0.13.2 but it becomes available in 0.13.3)

Issue #152, opened 2017-03-19
Height is not being verified for electrumx peers


shsmith:
We verify height is within 5 blocks for legacy peers, but this test is skipped for electrumx peers. 
This leads to the possibility of an electrumx peer with a deamon on a different chain being included in the peers. 
I think it is prudent to verify block height for all peers.

PR #153, opened 2017-03-19
fix https://github.com/kyuupichan/electrumx/issues/152

shsmith:
verify height of all peers, not just legacy peers to fix https://github.com/kyuupichan/electrumx/issues/152

### DIFF
--- a/server/peers.py
+++ b/server/peers.py
@@ -153,9 +153,9 @@ class PeerSession(JSONSession):
                 self.peer_verified(True)
                 self.peer.update_features(features)
                 verified = True
-        # For legacy peers not implementing features, check their height
+        # For all peers, check their height
         # as a proxy to determining they're on our network
-        if not verified and not self.peer.bad:
+        if not self.peer.bad:
             self.send_request(self.on_headers, 'blockchain.headers.subscribe')
         self.close_if_done()
 


erasmospunk:
utACK

Could you please `git rebase origin/master` so there is only one commit? The "update raspberrypi3 install procedure" commit is already in master.

bauerj:
And while you're at it, it's probably better to set the base branch to 1.1-prerelease.

shsmith:
I spoke with kyuupichan about this and think he plans to do it differently -- in a way that tags out-of-sync peers as stale rather than bad.

shsmith:
>Could you please git rebase origin/master so there is only one commit? The "update raspberrypi3 install procedure" commit is already in master.

That is exactly what I did, but I admit my git skills are minimal. 
I will try doing "git reset --hard upstream/master".

server/peers.py
```diff
@@ -153,9 +153,9 @@ def on_features(self, features, error):
                 self.peer_verified(True)
                 self.peer.update_features(features)
                 verified = True
-        # For legacy peers not implementing features, check their height
+        # For all peers, check their height
         # as a proxy to determining they're on our network
-        if not verified and not self.peer.bad:
+        if not self.peer.bad:
             self.send_request(self.on_headers, 'blockchain.headers.subscribe')
```
kyuupichan: I've removed the comment and these lines of code entirely.  Instead I've moved the send_request so it's unconditional into the connection_made() function.

PR #173, opened 2017-04-29
Allow multiple address versions for each coin

pooler:
Litecoin is going to transition to a new P2SH address version in the upcoming release. This patch allows ElectrumX to support both old and new version prefixes. For uniformity, all address version parameters are converted to lists, with the first entry representing the default version.


kyuupichan:
Thanks!  Let's just make P2PKH an array for now and leave the others.

PR #176, opened 2017-05-01
Allow custom Daemon and BlockProcessor classes

erasmospunk:
This is part of work in progress to support more coins that need extra control for the Daemon and BlockProcessor classes.

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -40,6 +40,8 @@ import lib.util as util
 from lib.hash import Base58, hash160, double_sha256, hash_to_str
 from lib.script import ScriptPubKey
 from lib.tx import Deserializer, DeserializerSegWit, DeserializerAuxPow, DeserializerZcash
+from server.block_processor import BlockProcessor
+from server.daemon import Daemon
 
 Block = namedtuple("Block", "header transactions")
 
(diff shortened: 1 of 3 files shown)


server/controller.py
```diff
@@ -50,8 +50,8 @@ def __init__(self, env):
         self.loop.set_default_executor(self.executor)
         self.start_time = time.time()
         self.coin = env.coin
-        self.daemon = Daemon(env.coin.daemon_urls(env.daemon_url))
-        self.bp = BlockProcessor(env, self, self.daemon)
+        self.daemon = self.coin.daemon()(env.coin.daemon_urls(env.daemon_url))
+        self.bp = self.coin.block_processor()(env, self, self.daemon)
```
kyuupichan: This means we can lose 2 imports

lib/coins.py
```diff
@@ -293,6 +295,15 @@ def electrum_header(cls, header, height):
...
         return Deserializer
 
+    @classmethod
+    def daemon(cls):
+        return Daemon
+
+    @classmethod
+    def block_processor(cls):
+        return BlockProcessor
+
```
kyuupichan: Any reason to have this a function and not a class member variable?  It can't change as the function takes no args to deduce a change.

erasmospunk: I prefer a class variable but did it in the same way as the:

```
def deserializer(cls):
        return Deserializer
```

If you like, I could refactor the `deserializer()` to be property as well.

server/controller.py
erasmospunk: Thanks, will clean up the imports in this file.

lib/coins.py
kyuupichan: Please. There it's a remnant of old code I think that did something different.

erasmospunk:
I did the deserializer refactor and cleaned up the imports in the controller.py. Didn't have the time to test the updated code yet.

server/controller.py
```diff
@@ -7,25 +7,22 @@
...
 from lib.hash import double_sha256, hash_to_str, hex_str_to_hash
+from lib.jsonrpc import JSONSessionBase, RPCError
 from lib.peer import Peer
-import lib.util as util
-from server.block_processor import BlockProcessor
-from server.daemon import Daemon, DaemonError
+from server.daemon import DaemonError
 from server.mempool import MemPool
 from server.peers import PeerManager
 from server.session import LocalRPC, ElectrumX
```
kyuupichan: John - I like to keep the imports split into 3 as per PEP8 I think - Python native imports first, 3rd party libs second, and then the app itself 3rd.  The change looks good apart from this, though I am unable to test it at the moment.
Can you restore the import order please and let me know if you manage to give it a test.  Thanks.

erasmospunk: Yeah, I used the import tidying of IntelliJ and it was too eager.

erasmospunk:
Fixed the imports in the controller.py. Note that I also removed all imports that we not needed (like `random`, `warnings`)

PR #177, opened 2017-05-03
Added SERVER_(SUB)VERSION to banner variables

laodc:
Add new variables to BANNER(_TOR)
$SERVER_VER will return the version number (eg: 1.0.10)
$SERVER_SUBVERSION will return the full version string (eg: ElectrumX 1.0.10)

$VERSION is kept for legacy which is the same as $SERVER_SUBVERSION


kyuupichan:
VERSION is used in several places in the codebase.  Let's not change VERSION.  Instead use version.split() like setup.py does.

kyuupichan:
In other words, let's keep this change localised to the banner function - just one file.

bauerj:
I'm curious. What's the use-case for `SERVER_VER`?

laodc:
Ok, i'll change it to isolated to banner only.

example banner:

 bitcoind-git: v$DAEMON_VERSION-55f641c
electrumX-git: v$SERVER_VERSION-2fe6793

laodc:
updated the code.

PR #182, opened 2017-05-14
Allow custom ElectrumX class

thelazier:


kyuupichan:
What is the idea here?  To add Dash-specific something?

thelazier:
Yes, I'm about to add https://github.com/thelazier/electrumx/commit/f179c679353cb38636f48cccd83d8e8afc3b36d5

PR #183, opened 2017-05-14
Add support for Dash Masternode methods

thelazier:
This will add Dash specific feature which require #182 to be merged so it's also include #182 .

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -572,6 +572,8 @@ class DogecoinTestnet(Dogecoin):
 
 # Source: https://github.com/dashpay/dash
 class Dash(Coin):
+    from server.session import DashElectrumX
+    from server.daemon import DashDaemon
     NAME = "Dash"
     SHORTNAME = "DASH"
     NET = "mainnet"
@@ -596,9 +598,7 @@ class Dash(Coin):
         'electrum.dash.siampm.com s t',
         'wl4sfwq2hwxnodof.onion s t',
     ]
-    from server.session import DashElectrumX
     SESSIONCLS = DashElectrumX
-    from server.daemon import DashDaemon
     DAEMON = DashDaemon
 
     @classmethod
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -596,6 +596,10 @@ class Dash(Coin):
         'electrum.dash.siampm.com s t',
         'wl4sfwq2hwxnodof.onion s t',
     ]
+    from server.session import DashElectrumX
+    SESSIONCLS = DashElectrumX
+    from server.daemon import DashDaemon
+    DAEMON = DashDaemon
 
     @classmethod
     def header_hash(cls, header):
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -42,6 +42,7 @@ from lib.script import ScriptPubKey
 from lib.tx import Deserializer, DeserializerSegWit, DeserializerAuxPow, DeserializerZcash
 from server.block_processor import BlockProcessor
 from server.daemon import Daemon
+from server.session import ElectrumX
 
 Block = namedtuple("Block", "header transactions")
 
@@ -61,6 +62,7 @@ class Coin(object):
     HASHX_LEN = 11
     BASIC_HEADER_SIZE = 80
     STATIC_BLOCK_HEADERS = True
+    SESSIONCLS = ElectrumX
     DESERIALIZER = Deserializer
     DAEMON = Daemon
     BLOCK_PROCESSOR = BlockProcessor
--- a/server/session.py
+++ b/server/session.py
@@ -415,7 +415,8 @@ class DashElectrumX(ElectrumX):
 
     def server_version(self, client_name=None, protocol_version=None):
         '''Returns the server version as a string.
-        Force version string response for Electrum-Dash 2.6.4
+        Force version string response for Electrum-Dash 2.6.4 client caused by
+        https://github.com/dashpay/electrum-dash/commit/638cf6c0aeb7be14a85ad98f873791cb7b49ee29
         '''
 
         default_return = super().server_version(client_name, protocol_version)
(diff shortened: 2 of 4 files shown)


lib/coins.py
```diff
@@ -594,6 +596,10 @@ class Dash(Coin):
         'electrum.dash.siampm.com s t',
         'wl4sfwq2hwxnodof.onion s t',
     ]
+    from server.session import DashElectrumX
+    SESSIONCLS = DashElectrumX
+    from server.daemon import DashDaemon
+    DAEMON = DashDaemon
```
kyuupichan: Just put these imports at the top please

server/session.py
```diff
@@ -387,3 +387,69 @@ def __init__(self, *args, **kwargs):
...
+                    'method': 'masternode.subscribe',
+                    'params': [masternode],
+                    'result': status.get(masternode),
+                }
+                self.send_binary(self.encode_payload(payload))
+
+    def server_version(self, client_name=None, protocol_version=None):
+        '''Returns the server version as a string.
+        ForcE version string response for Electrum-Dash 2.6.4
+        '''
```
kyuupichan: ForcE typo.  Can you also please add a comment explaining why the special case is needed.

server/session.py
```diff
@@ -387,3 +387,69 @@ def __init__(self, *args, **kwargs):
...
+    # Masternode command handlers
+    async def masternode_announce_broadcast(self, signmnb):
+        '''Pass through the parameters to the daemon.
+
+        An ugly API: current Electrum clients only pass the masternode
+        broadcast message in hex and expect error messages to be returned in
+        the result field.  And the server shouldn't be doing the client's
+        user interface job here.
+        '''
+        try:
```
kyuupichan: Is there a plan to fix this API?

server/session.py
```diff
@@ -387,3 +387,69 @@ def __init__(self, *args, **kwargs):
...
+
+
+class DashElectrumX(ElectrumX):
+    '''A TCP server that handles incoming Electrum Dash connections.'''
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.electrumx_handlers['masternode.announce.broadcast'] = self.masternode_announce_broadcast
+        self.electrumx_handlers['masternode.subscribe'] = self.masternode_subscribe
+        self.subscribe_mns = False
```
kyuupichan: I don't think we need this boolean variable.  You can just loop over the mns set.

server/session.py
```diff
@@ -387,3 +387,69 @@ def __init__(self, *args, **kwargs):
...
+        self.mns = set()
+
+    async def notify(self, height, touched):
+        '''Notify the client about changes in masternode list.'''
+
+        await super().notify(height, touched)
+
+        if self.subscribe_mns:
+            for masternode in self.mns:
```
kyuupichan: The condition should not be necessary right?

server/session.py
```diff
@@ -387,3 +387,69 @@ def __init__(self, *args, **kwargs):
...
+            self.log_info('masternode_broadcast: {}'.format(message))
+            return (
+                'The masternode broadcast was rejected.  ({})\n[{}]'
+                .format(message, signmnb)
+            )
+
+    async def masternode_subscribe(self, vin):
+        '''Returns the status of masternode.'''
+        result = await self.daemon.masternode_list(['status',vin])
+        if result is not None:
```
kyuupichan: Space after comma

server/session.py
```diff
@@ -387,3 +387,69 @@ def __init__(self, *args, **kwargs):
...
+            self.mns.add(vin)
+            self.subscribe_mns = True
+        return result.get(vin)
```
kyuupichan: If result can be None we cannot "get" it.  If it cannot be None we don't need to test.

lib/coins.py
thelazier: Top Lines inside `class Dash(Coin):` , right?

server/session.py
thelazier: I think this API won't be fixed, it should be similar to broadcast transaction and the daemon would do all the verified by network anyway. So, Comment is removed.

lib/coins.py
thelazier: Done via 	8d21eae

server/session.py
thelazier: Agree.

server/session.py
thelazier: Agree.

server/session.py
thelazier: Typo is fixed.

server/session.py
thelazier: Yes, sir.

server/session.py
thelazier: You are correct.

server/session.py
```diff
@@ -387,3 +387,62 @@ def __init__(self, *args, **kwargs):
...
+                'method': 'masternode.subscribe',
+                'params': [masternode],
+                'result': status.get(masternode),
+            }
+            self.send_binary(self.encode_payload(payload))
+
+    def server_version(self, client_name=None, protocol_version=None):
+        '''Returns the server version as a string.
+        Force version string response for Electrum-Dash 2.6.4
+        '''
```
kyuupichan: Please comment on why the special case

thelazier: This is directly related to https://github.com/dashpay/electrum-dash/commit/638cf6c0aeb7be14a85ad98f873791cb7b49ee29 so this special case will be removed later when newer version of electrum-dash is released without this server version checking.

kyuupichan:
I think the only thing missing is the move of the imports to the top of the file (not function) that I requested a couple of times.

thelazier:
Moved and Pushed as https://github.com/thelazier/electrumx/commit/50c0cf66b1c897532c6002c92bc17fde7a030805

kyuupichan:
Any reason you've not repushed as this PR so I can just click Merge?

thelazier:
I pushed in same branch with this PR but I don't know why the PR is not update

kyuupichan:
OK please fix that in any follow-ups you might have.

PR #184, opened 2017-05-15
Pass information about the latest client version to the client

bauerj:
This is based on this IRC conversation:

```
> [11:23:56] <skace> ThomasV: any way to allow proxy settings for 'auto update' feature... or... make it a check box "check for new updates" yes/no
> [11:24:04] <skace> i hate having to check the site if something new
> [11:25:11] <ThomasV> skace: I think we need to do it so that the client checks from the electrum server, instead of requesting directly from electrum.org
> [11:25:22] <ThomasV> that way there's no privacy loss
> [11:25:43] <ThomasV> the way it used to work was direct https request
> [11:26:35] <skace> so server would have to query website?
> [11:26:42] <skace> i would of assumed proxy settings would be enough
> [11:32:34] <ThomasV> not everyone uses a proxy
> [11:32:54] <ThomasV> it's better if the server queries the website, I think
> [11:36:52] <EagleTM> just an idea: could be made a server feature in e-x? server queries website periodically, gives info to client.
> [11:37:09] <ThomasV> EagleTM: yes, that's the idea
> [11:37:25] <EagleTM> ah ok ;) yes just re-read it
```

So the idea is that the server regularly checks the website for the latest client version. The client would then get that information from the server and notify the user if something newer is available.

I'm not sure if the `Coin` class is the best place for the `latest_version` method. Maybe it would make more sense to move that to the `Session` class?

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -72,6 +72,7 @@ class Coin(object):
     # Peer discovery
     PEER_DEFAULT_PORTS = {'t': '50001', 's': '50002'}
     PEERS = []
+    LATEST_CLIENT_VERSION_URL = None
 
     @classmethod
     def lookup_coin_class(cls, name, net):
@@ -332,6 +333,7 @@ class Bitcoin(Coin):
     IRC_PREFIX = "E_"
     IRC_CHANNEL = "#electrum"
     RPC_PORT = 8332
+    LATEST_CLIENT_VERSION_URL = "https://electrum.org/version"
     PEERS = [
         'btc.smsys.me s995',
         'electrum.be s t',
--- a/lib/util.py
+++ b/lib/util.py
@@ -35,6 +35,8 @@ import re
 import sys
 from collections import Container, Mapping
 
+import time
+
 
 class LoggedClass(object):
 
@@ -62,11 +64,12 @@ class LoggedClass(object):
         self.logger.error(self.log_prefix + msg)
 
 
-# Method decorator.  To be used for calculations that will always
-# deliver the same result.  The method cannot take any arguments
-# and should be accessed as an attribute.
 class cachedproperty(object):
-
+    '''
+    Method decorator.
+    The result of the decorated method will be cached forever and
+    will be available as an attribute with the same name as the decorated method.
+    '''
     def __init__(self, f):
         self.f = f
 
@@ -76,6 +79,31 @@ class cachedproperty(object):
         setattr(obj, self.f.__name__, value)
         return value
 
+
+class expiring_cachedproperty(object):
+    '''
+    Method decorator.
+    The result of the decorated method will be cached for `expire_time` seconds
+    and will be available as an attribute with the same name as the decorated
+    method.
+    '''
+    def __init__(self, expire_time):
+        self.expire_time = expire_time
+        self.last_call = 0
+
+    def __call__(self, f):
+        self.f = f
+        return self
+
+    def __get__(self, obj, type):
+        if hasattr(obj, "_cached_" + self.f.__name__) and self.last_call + self.expire_time > time.time():
+            return getattr(obj, "_cached_" + self.f.__name__)
+        obj = obj or type
+        value = self.f(obj)
+
+        self.last_call = time.time()
+        setattr(obj, "_cached_" + self.f.__name__, value)
+        return value
 
 def formatted_time(t, sep=' '):
     '''Return a number of seconds as a string in days, hours, mins and
--- a/server/controller.py
+++ b/server/controller.py
@@ -16,6 +16,7 @@ from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 
+import aiohttp
 import pylru
 
 from lib.jsonrpc import JSONSessionBase, RPCError
@@ -871,3 +872,15 @@ class Controller(util.LoggedClass):
     def donation_address(self):
         '''Return the donation address as a string, empty if there is none.'''
         return self.env.donation_address
+
+    @util.expiring_cachedproperty(86400)
+    async def latest_version(self):
+        if not self.coin.LATEST_CLIENT_VERSION_URL:
+            return None
+        async with aiohttp.ClientSession() as session:
+            async with session.get(self.coin.LATEST_CLIENT_VERSION_URL) as response:
+                if response.status == 200:
+                    return await response.text()
+
+
+
--- a/server/session.py
+++ b/server/session.py
@@ -125,6 +125,7 @@ class ElectrumX(SessionBase):
             'server.features': self.server_features,
             'server.peers.subscribe': self.peers_subscribe,
             'server.version': self.server_version,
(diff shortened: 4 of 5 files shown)


lib/util.py
```diff
@@ -62,18 +64,34 @@ def log_error(self, msg):
...
 class cachedproperty(object):
-
-    def __init__(self, f):
+    '''
+    Method decorator.
+    The result of the decorated method will be cached forever or for
+    `expire_time` seconds and will be available as an attribute with the same
+    name as the decorated method.
+    '''
+    def __init__(self, *args, expire_time=None):
```
kyuupichan: Can you create a new decorator entirely for time-cached properties?  I'd like to keep the existing one as-is.  Nice idea, b.t.w.

lib/coins.py
```diff
@@ -295,6 +297,16 @@ def electrum_header(cls, header, height):
...
+    @util.cachedproperty(expire_time=86400)
+    async def latest_version(cls):
+        if not hasattr(cls, "UPDATECHECK_URL"):
+            return None
+        async with aiohttp.ClientSession() as session:
+            async with session.get(cls.UPDATECHECK_URL) as response:
+                if response.status == 200:
+                    return await response.text()
+
+
```
kyuupichan: Can you change the variable name to LATEST_CLIENT_VERSION_URL and have the base class set to None, overriding in the Bitcoin class?

The cached property and aiohttp logic should be somewhere else; probably in controller.py as there is a single instance of that class.

lib/util.py
bauerj: Sure and thanks 👍

bauerj:
Well, it seems to work only once now:

	May 16 20:40:34 helicarrier electrumx_server.py[10535]:   File "/mnt/ultra.bauerj/electrumx/lib/jsonrpc.py", line 528, in process_single_request
	May 16 20:40:34 helicarrier electrumx_server.py[10535]:     result = await self.handle_payload(payload, self.request_handler)
	May 16 20:40:34 helicarrier electrumx_server.py[10535]:   File "/mnt/ultra.bauerj/electrumx/lib/jsonrpc.py", line 608, in handle_payload
	May 16 20:40:34 helicarrier electrumx_server.py[10535]:     return await handler(**kw_args)
	May 16 20:40:34 helicarrier electrumx_server.py[10535]:   File "/mnt/ultra.bauerj/electrumx/server/session.py", line 342, in client_latest_version
	May 16 20:40:34 helicarrier electrumx_server.py[10535]:     v = await self.controller.latest_version
    May 16 20:41:34 helicarrier electrumx_server.py[10535]: RuntimeError: cannot reuse already awaited coroutine

Not sure what to do there 😕

kyuupichan:
I don't think these kind of decorators will work well with async routines :(

erasmospunk:
NACK, there are other, beyond electrum.org, clients and services that use the electrum protocol/server.

A better approach is to use a plugin that can be installed by the server operator.

bauerj:
Why shouldn't this be part of electrumx? Other client forks could use it as well.

erasmospunk:
No, I am talking about other bitcoin wallets like Coinomi. There is no benefit to the server to support this feature and limit itself to a single wallet.

Let's keep the protocol clean and simple.

bauerj:
> There is no benefit to the server to support this feature...

In a way, there is no benefit for the server to support any feature 😉. The benefits are all client-side.

> ... and limit itself to a single wallet.

This wouldn't keep other wallets from using the Electrum protocol, they don't have to use this feature.

> Let's keep the protocol clean and simple.

I agree that this should be our goal - although I believe that the current protocol isn't perfectly clean either. Do you see a better way to implement this?

bauerj:
I wonder why it works with bauerj/electrumx@6b08a527951a58f0cdec8fb55614d8f65b5fabde but not with bauerj/electrumx@7e0036a 🤔

erasmospunk:
Why do you want to add a vendor specific client version check in a server that is currently used by other clients as well?

There is a very clean way to implement this, each client asks their own distribution site for their version like they do now. Why this PR is better than simply contacting the website via HTTPS (and TOR if you want privacy)?

bauerj:
> Why this PR is better than simply contacting the website via HTTPS (and TOR if you want privacy)?

The only reason is that Tor is an external dependency that the user needs to take care of while an electrum server is always available.

erasmospunk:
A user that cares about their privacy would not connect via the clear net to a public electrum server.

Also you haven't answered about the possibility of having alternative client vendors. Shouldn't they also have a way to query their version codes via the electrum protocol?

Consider adding a parameter to get the version for a particular vendor i.e.:

```
{"method":"client.latest_version", "params": ["org.electrum"]}
```
Also, please make it configurable to disable this feature completely.

That being said, I still don't understand why this feature is useful: Yeah, there is a narrow use case for some privacy but still you have to download the new version from electrum.org so you lose any benefit. Furthermore if you didn't want the privacy leak, you would just disable the automatic version check, right? This doesn't justify the LOC increase.

bauerj:
> A user that cares about their privacy would not connect via the clear net to a public electrum server.

I, for example, don't really care about the loss of privacy caused by connecting to my own server through the clearnet. I wouldn't care about electrum.org having my IP address either though. My point is that there are probably more increments between "user doesn't care about privacy" and "user cares about privacy".

> Also you haven't answered about the possibility of having alternative client vendors. Shouldn't they also have a way to query their version codes via the electrum protocol?

Sure, if there is any interest.

> Consider adding a parameter to get the version for a particular vendor

That sounds like a good idea!

> Also, please make it configurable to disable this feature completely.

Sure, I'll add that but I suggest that it would be enabled by default.

> That being said, I still don't understand why this feature is useful: Yeah, there is a narrow use case for some privacy but still you have to download the new version from electrum.org so you lose any benefit. 

Consider a client-side check that happens every time you start Electrum. If the information would be fetched directly from their server, they would be able to tell precisely when you started Electrum.

> Furthermore if you didn't want the privacy leak, you would just disable the automatic version check, right? 

That's true but the default settings should already prevent from such a privacy leak.

> This doesn't justify the LOC increase.

I'm not sure if there's an objective way to tell that.

kyuupichan:
I thought this was a good idea but JJ has a point I think.  It seems this is designed to allay negative points from privacy project reviews that don't actually make a whole lot of sense when viewed holistically as JJ has been pointing out.  Perhaps this is better client-side after all.

bauerj:
Okay, I agree that there are simpler solutions to this problem (especially since this one is not even working).

EagleTM:
This is already closed but still I'd give it a Concept ACK
Besides the privacy features, it's a caching layer, we don't have to scale the electrum.org webservers to deal with requests for new versions.

I believe Thomas has in mind to have the server also send other useful info like exchange rate data and eventually use it for LN things, so warming to the idea of having a more feature rich server is not so bad.

followtheart:
Does electrumx compatible with coinomi？ @erasmospunk

PR #188, opened 2017-05-18
Remove dead code

erasmospunk:


kyuupichan:
These are used in year-old unreleased code, but that likely requires rewriting too, so sure.

PR #194, opened 2017-06-20
Add Dockerfile

followtheart:
Add Dockerfile reference

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,8 +18,11 @@ Getting Started
 
 See `docs/HOWTO.rst`_.
 There is also an `installer`_ available that simplifies the installation on various Linux-based distributions.
+There is also an `Dockerfile`_ available .
 
 .. _installer: https://github.com/bauerj/electrumx-installer
+
+.. _Dockerfile: https://github.com/followtheart/electrumx-docker
 
 Features
 ========


followtheart:
If electrumx need a dockerfile in source ,I will submit another pr ,with dockerfile merged.

bauerj:
Your PR seems to be fine but I have some general questions about your docker image:

- Is it possible to set `RLIMIT_NOFILE` (`ulimit -n`) to a higher value within the docker container?
- Wouldn't it make more sense to mount the database into a volume so you wouldn't have to rebuild the database when you update the container?
- Can you release it to the Docker store?
- Is there a particular reason to use 14.04 as the base image?

followtheart:
1.Since setting ulimit settings in a container requires extra privileges not available in the default container, you can set these using the --ulimit flag.See [docker here](https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container-ulimit)
2.It good idea to storage data into volumes,dockerfile have updated to fix this(/db for leveldb,/log for logs).
3.Now ,It has been released to docker hub.
4.No particular reasons to use 14.04,just my ower server version :haircut_man: ,now using latest(16.04).

bauerj:
Perfect, thanks! 

The only thing that seems a bit odd is to spawn a shell in the container to edit the configuration files. A more commonly used method I saw with other images is to either mount the configuration directory from the host or to use environment variables (like this: https://dantehranian.wordpress.com/2015/03/25/how-should-i-get-application-configuration-into-my-docker-containers/). Do you know if that would make sense with your image?

Anyway, thanks for your work so far 👍

followtheart:
Its a good practice to mount the configuration directory from the host.And now is updated,using a volume `/env` ,you can mount with your local configuration.

Thx,for your suggestions.

PR #197, opened 2017-06-24
fix LegacyRPCDaemon bag

namuyan:
I find a bag on LegacyRPCDaemon class.
It's not False because getblock with False flag export blockhash as HEX code.


kyuupichan:
@erasmospunk - is this correct?  How did the existing code work for you?

namuyan:
Old type coinds works because they aren't affected by True or False flag.
But in the future, the flag is enabled, it will be bug, I think.
I suggest correct state or removing the flag.

PR #205, opened 2017-07-11
Improve daemon JSON RPC compatibility

erasmospunk:
This commit adds an id counter as per the JSON RPC specs. Additionally
it is possible for a Daemon subclass to specify a custom transport
layer.

### DIFF
--- a/server/daemon.py
+++ b/server/daemon.py
@@ -45,6 +45,7 @@ class Daemon(util.LoggedClass):
         self.workqueue_semaphore = asyncio.Semaphore(value=10)
         self.down = False
         self.last_error_time = 0
+        self.req_id = 0
         # assignment of asyncio.TimeoutError are essentially ignored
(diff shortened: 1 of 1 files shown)


kyuupichan:
John - can you explain what motivates these changes?

Also, is it possible to add some kind of tests for these?

erasmospunk:
I was trying to make it work with [btcd](https://github.com/btcsuite/btcd) and noticed the missing id parameter and the inability to make ElectrumX work with HTTPS RPCs. This pull req solves some of the issues.

I would liked to add tests but it's not clear to me how to proceed in the case of the Daemon as it needs some kind of a mock server.

kyuupichan:
OK thanks John; I'll push a slightly different version but that should be good enough for what you want.

erasmospunk:
Looks good, thanks!

PR #206, opened 2017-07-18
Update HOWTO.rst

zhovner:


thokon00:
Trivial, but **Mac OS X** seems to be a rather common notation.

bauerj:
AFAIK it's now called macOS. OS X was the old name.

thokon00:
Ha, indeed they did. Ignore my comment then.

PR #207, opened 2017-07-19
update FairCoin genesis block

thokon00:
The new version of FairCoin has been launched yesterday. This is the new genesis block and the current server list.


kyuupichan:
Huh?  What happened to the old one?

thokon00:
FairCoin1 has stopped and all balances have been transferred to the new blockchain on the 18th of July. FairCoin2 features proof-of-cooperation instead of PoS in FairCoin1. ElectrumX has never been ported to FairCoin1 but only to FairCoin2 beta. Now we're live and thus the genesis block is now final.

More information can be found here: [FairCoin2 migration FAQ](https://github.com/faircoin/faircoin/wiki/FairCoin2-migration-FAQ)

kyuupichan:
OK, I'll take your word for it :)

PR #211, opened 2017-08-02
Update controller.py

gaoyaxing24:
I have a problem in version 1.0.15. 
localrpc error info 'ERROR:Controller:RPC server failed to listen on localhost:8000 :[Errno 99] error while attempting to bind on address ('::1', 8000, 0, 0): cannot assign requested address' If you starting service. 
when you replace 'localhost' to '127.0.0.1' in control.py . it is OK! 
bacause someome don't set localhost:127.0.0.1 in host or the system cannot parse
I've found this problem on my 3 servers.

### DIFF
--- a/server/controller.py
+++ b/server/controller.py
@@ -198,7 +198,7 @@ class Controller(util.LoggedClass):
     async def main_loop(self):
         '''Controller main loop.'''
         if self.env.rpc_port is not None:
-            await self.start_server('RPC', 'localhost', self.env.rpc_port)
+            await self.start_server('RPC', '127.0.0.1', self.env.rpc_port)
         self.ensure_future(self.bp.main_loop())
         self.ensure_future(self.wait_for_bp_catchup())
 


kyuupichan:
Can you tell me if None works as the 2nd argument instead?

gaoyaxing24:
It is working normal when I try replace '127.0.0.1' or 'localhost' to None.
But I suggested use explicit var or string. Not None. Because it's easier to understand.

kyuupichan:
Someone else had a different issue.  None is best because it's documented to mean all interfaces including IPv6.  Do you want to update this PR or shall I make the change?

gaoyaxing24:
Your are right! It's really something I didn't take into account.
I think shall you make the change. Because I'm not quite familiar with the code yet.

kyuupichan:
OK, I'll do it now.  Thanks for testing - I didn't make the change as I don't have the issue so had no way to test.

gaoyaxing24:
OK! you can find me anytime if you need test or else.

PR #214, opened 2017-08-06
Minor optimizations to Bitcoin tx processing

JustinTArthur:
Reduces a small amount of overhead from:
* Object attribute referencing (`LOAD_ATTR`)
* `unpack_from` format processing
* Multiple `len()` calls

### DIFF
--- a/lib/tx.py
+++ b/lib/tx.py
@@ -29,10 +29,11 @@
 
 
 from collections import namedtuple
-from struct import unpack_from
 
-from lib.util import cachedproperty
 from lib.hash import double_sha256, hash_to_str
+from lib.util import (cachedproperty, unpack_int32_from, unpack_int64_from,
+                      unpack_uint16_from, unpack_uint32_from,
+                      unpack_uint64_from)
 
 
 class Tx(namedtuple("Tx", "version inputs outputs locktime")):
@@ -149,27 +150,27 @@ class Deserializer(object):
         return self._read_le_uint64()
 
     def _read_le_int32(self):
-        result, = unpack_from('<i', self.binary, self.cursor)
+        result, = unpack_int32_from(self.binary, self.cursor)
         self.cursor += 4
         return result
 
     def _read_le_int64(self):
-        result, = unpack_from('<q', self.binary, self.cursor)
+        result, = unpack_int64_from(self.binary, self.cursor)
         self.cursor += 8
         return result
 
     def _read_le_uint16(self):
-        result, = unpack_from('<H', self.binary, self.cursor)
+        result, = unpack_uint16_from(self.binary, self.cursor)
         self.cursor += 2
         return result
 
     def _read_le_uint32(self):
-        result, = unpack_from('<I', self.binary, self.cursor)
+        result, = unpack_uint32_from(self.binary, self.cursor)
         self.cursor += 4
         return result
 
     def _read_le_uint64(self):
-        result, = unpack_from('<Q', self.binary, self.cursor)
+        result, = unpack_uint64_from(self.binary, self.cursor)
         self.cursor += 8
         return result
 
--- a/lib/tx.py
+++ b/lib/tx.py
@@ -79,6 +79,7 @@ class Deserializer(object):
     def __init__(self, binary, start=0):
         assert isinstance(binary, bytes)
         self.binary = binary
+        self.binary_length = len(binary)
         self.cursor = start
 
     def read_tx(self):
@@ -132,7 +133,7 @@ class Deserializer(object):
     def _read_nbytes(self, n):
         cursor = self.cursor
         self.cursor = end = cursor + n
-        assert len(self.binary) >= end
+        assert self.binary_length >= end
         return self.binary[cursor:end]
 
     def _read_varbytes(self):
--- a/docs/PERFORMANCE-NOTES
+++ b/docs/PERFORMANCE-NOTES
@@ -1,4 +1,4 @@
-Just some notes on performance with Python 3.5.  I am taking this into
+Just some notes on performance with Python 3.5. We are taking these into
 account in the code.
 
 - 60% faster to create lists with [] list comprehensions than tuples
@@ -16,7 +16,14 @@ account in the code.
 
 - struct.pack, struct.unpack are over 60% faster than int.to_bytes and
   int.from_bytes.  They are faster little endian (presumably because
-  it matches the host) than big endian regardless of length.
+  it matches the host) than big endian regardless of length. Furthermore,
+  using stored packing and unpacking methods from Struct classes is faster
+  than using the flexible-format struct.[un]pack equivalents.
+
+  After storing the Struct('<Q').unpack_from function as unpack_uint64_from,
+   later calls to unpack_uint64_from(b, 0) are about 30% faster than calls to
+   unpack_from('<Q', b, 0).
+
 
 - single-item list and tuple unpacking.  Suppose b = (1, )
 
@@ -24,3 +31,6 @@ account in the code.
            and about 45% faster than a = b[0]
 
 - multiple assignment is faster using tuples only for 3 or more items
+
+- retrieving a previously stored length of a bytes object can be over 200%
+  faster than a new call to len(b)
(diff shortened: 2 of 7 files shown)


kyuupichan:
Have you measured any timings?  This needs tests for all new lib/ APIs.

JustinTArthur:
I have. Here is a simplified-for-benchmarking version of these optimizations to be run in an electrumx Python environment:
```python
from timeit import timeit

iterations = 1000000


def run_strategies(strategies, override_iterations = None, context=None):
    for n, strategy in enumerate(strategies):
        title, setup, stmt = strategy
        is_new_method = n % 2
        if is_new_method:
            prior_result = result

        result = timeit(stmt=stmt, setup=setup, number=override_iterations or iterations, globals=context)

        if is_new_method:
            if prior_result < result:
                speedup = "{}x slower".format(result/prior_result)
            else:
                speedup = "{}x faster".format(prior_result/result)
            print("\t{}:\t{:.6f}ms ({})".format(title, result*1000, speedup))
        else:
            print("\t{}:\t{:.6f}ms".format(title, result*1000))
    print("\n")

print("len() performance".format(iterations))
small_block_setup = "binary = bytes(i % 255 for i in range(1024))\nend=700"
large_block_setup = "binary = bytes(i % 255 for i in range(1024*1024))\nend=700"
preassignment_setup = "binary_length = len(binary)"

strategies = (
    ('len() with small block', small_block_setup, 'assert len(binary) >= end'),
    ('Preassigned length with small block', small_block_setup + "\n" + preassignment_setup, 'assert binary_length >= end'),

    ('len() with large block', large_block_setup, 'assert len(binary) >= end'),
    ('Preassigned length with large block', large_block_setup + "\n"+ preassignment_setup, 'assert binary_length >= end'),
)
run_strategies(strategies)


print("hash method referencing performance".format(iterations))
strategies = (
    ("Dot reference every time", "import hashlib\nbinary=bytes(i % 255 for i in range(1024))", "hashlib.sha256(binary).digest()"),
    ("Dot reference ahead of time", "import hashlib\nsha256=hashlib.sha256\nbinary=bytes(i % 255 for i in range(1024))", "sha256(binary).digest()")
)
run_strategies(strategies)

print("hexing/unhexing strategies".format(iterations))
strategies = (
    ("hexlify", "from binascii import hexlify\nbinary = bytes(i % 255 for i in range(1024))", "hexlify(binary)"),
    ("bytes.hex", "binary = bytes(i % 255 for i in range(1024))", "binary.hex()"),

    ("bytes.fromhex", "h = 'ABCDEF1234567890ABCDEF'", "bytes.fromhex(h)"),
    ("unhexlify", "from binascii import unhexlify\nh = 'ABCDEF1234567890ABCDEF'", "unhexlify(h)"),

    ("Always bytes.fromhex", "h = 'ABCDEF1234567890ABCDEF'", "bytes.fromhex(h)"),
    ("Pre-referenced bytes.fromhex", "h = 'ABCDEF1234567890ABCDEF'\nfromhex=bytes.fromhex", "fromhex(h)")
)
run_strategies(strategies)

print("unpacking")
strategies = (
    ("expanded", "from struct import unpack_from\nb=b'\\xb8\\x0cT\\x84\\xf2gv\\xbb'", "unpack_from('<Q', b, 0)"),
    ("classfull", "from lib.util import unpack_uint64_from\nb=b'\\xb8\\x0cT\\x84\\xf2gv\\xbb'", "unpack_uint64_from(b, 0)")
)
run_strategies(strategies)
```

Sample results on CPython 3.6.0:
```
len() performance
	len() with small block:	111.494676ms
	Preassigned length with small block:	28.924226ms (3.8547159751741455x faster)
	len() with large block:	111.240188ms
	Preassigned length with large block:	28.949783ms (3.8425223434608697x faster)


hash method referencing performance
	Dot reference every time:	4357.400288ms
	Dot reference ahead of time:	4237.259080ms (1.0283535195107778x faster)


hexing/unhexing strategies
	hexlify:	1918.105978ms
	bytes.hex:	1787.958140ms (1.0727913227349626x faster)
	bytes.fromhex:	184.133853ms
	unhexlify:	159.359805ms (1.155459828754998x faster)
	Always bytes.fromhex:	192.917301ms
	Pre-referenced bytes.fromhex:	123.772461ms (1.5586447861370305x faster)


unpacking
	expanded:	262.429034ms
	classfull:	201.242915ms (1.304041108632736x faster)
```

Do you need me to to produce unit tests for areas not previously covered by the current tests assuming these changes affect those areas?

JustinTArthur:
* I updated the branch to include unit tests of the newly exposed `lib` package members
  * They are mostly just references to what was already being used.
* I also renamed the `hashlib` members that are internal to the `lib.hash` module to have the underscore prefix.
  * They were already incorporated in the hash module unit tests.
* For consistency, I made even non-performance-critical code use the `hex_to_bytes` function.
* I added some notes to the performance doc.

bauerj:
Is there a significant difference in time needed to create the database?

JustinTArthur:
@bauerj statistically significant? Yes. User experience significant? Eh, maybe? On my beefy Mac Pro these changes give me about 1000 extra txes/s in the log output and cut down my from-genesis-to-today estimate from around _11h 35m_ to _10h 11m_.

These changes are taking some of the lowest hanging fruit from some of the hardest hit sections if you look at a profiler. I'm exploring some changes that are either harder to implement or harder to test that would make a more obvious dent. In no particular order, those are:
* **Use unix domain socket instead of TCP loopback to a local full node's RPC port**
  * TCP is sloooow
  * This is actually an easy addition, but we're probably going to have to wait until Core 0.16 depending on the libevent project's roadmap. Tracked in bitcoin/bitcoin#9919
* **uvloop option for remote RPC**
  * Not sure how involved this will be.
  * My only concern is if it will be compatible out of the box with the custom executor we use in controller.
* **Explore using a `memoryview` for slicing intensive operations.**
  * The unfortunate reality with `memoryview` is that after slicing things up, you eventually need to retrieve those bytes for hashing and other operations. This makes it less useful to test the optimization on a statement by statement basis. You need to test it's effect from start to finish.
* **Explore using `bytesarray` or `deque` for append-intensive operations.**
  * The tx processor for segregated witness on litecoin and bitcoin would be a good candidate for experimentation

I wouldn't be surprised if @kyuupichan has already been exploring a couple of these.

bauerj:
> *11h 35m to 10h 11m.*

That **is** significant!

lib/tx.py
```diff
@@ -29,9 +29,9 @@
 
 
 from collections import namedtuple
-from struct import unpack_from
 
-from lib.util import cachedproperty
+from lib.util import cachedproperty, unpack_int32_from, unpack_int64_from, unpack_uint16_from, unpack_uint32_from, \
+    unpack_uint64_from
```
kyuupichan: Please use parenthesis rather than backslash and restrict to 80 columns.

lib/tx.py
```diff
@@ -78,6 +78,7 @@ class Deserializer(object):
     def __init__(self, binary, start=0):
         assert isinstance(binary, bytes)
         self.binary = binary
+        self.binary_length = len(binary)
```
kyuupichan: I'm not a fan of this one and from your timings the effect is marginal compared to the others.  Can we drop the len() optimisation please.

kyuupichan:
@JustinTArthur thanks for the timing info, that is indeed fantastic. 

I'll merge once you fix the col wrap.  I would also like to drop the len speedup; it's marginal and IMO not worth it.  Thanks.

lib/tx.py
JustinTArthur: I would recommend leaving this change in. While `len()` is nice and explicit here, those timings show over 200% speedup for referencing the previous stored length over calling `len()` again. While its only shaving off tens of milliseconds, those add up over the many times the code is reached. Profiling before the change has it taking up almost 1% of my sync time: 
image

JustinTArthur:
I adjusted the import format using isort. As for the `len()` I attached a profile sample to the review. Maybe you are looking at the wrong thing in my timings? I still believe it is worth changing, but if you disagree, I will remove it from the PR.

kyuupichan:
Hey Justin, let's discuss next you're on IRC.

PR #215, opened 2017-08-07
Allow setting an alternate asyncio event loop policy (e.g. uvloop)

JustinTArthur:
I wanted the option to use uvloop. I didn't notice a significant difference networking on the loopback interface, but it's possible some will find it faster than the unix select loop that comes with CPython.

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -102,6 +102,15 @@ These environment variables are optional:
   to blank.  The default is appropriate for **COIN** and **NET**
   (e.g., 8000 for Bitcoin mainnet) if not set.
 
+* **EVENT_LOOP_POLICY**
+
+  Python path to an asyncio event loop policy class. The policy determines
+  what event loop implementation will be used to initiate and respond to
+  networking events.
+
+  For example, this can be used to employ uvloop if it is installed by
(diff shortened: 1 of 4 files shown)


kyuupichan:
Hi Justin,

I don't have anything against supporting other event loop policies.  However I don't like using arbitrary Python paths.  How about just accepting a well-known string, perhaps "asyncio" and "uvloop" for now?  And I think we should set the policy as a member function of Env, possibly called at the end of Env __init__ (though I won't insist on that), so that other top-level programs using the Env object can easily adopt the policy, either automatically or with a one-line function call.

What do you think?

JustinTArthur:
I'm ok with fixed symbols for loop implementations. Maybe blank/empty/missing for the platform default?

As for putting the policy execution in the Env class, I'll leave the decision up to you, as you know more than I where you want to go with the project architecture. It means `Env` instances will be more than just a data access objects, for what that's worth. Up until this point, the `Controller` was responsible for realizing the environment changes suggested in the Env DAO. Since loop instantiation happens prior to the Controller's setup, we do need *some* place to put it.

kyuupichan:
Justin - I've pushed a different change that accepts "uvloop" as a policy using your envvar.  Can you confirm it works as expected please.

PR #218, opened 2017-08-12
Add Fujicoin

fujicoin:
Hi bauerj,

I redoed the fork and made a new pull request.
I think that it was hard to understand because the time before fork was old.

Thanks,
motty

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -42,7 +42,7 @@ from lib.script import ScriptPubKey
 from lib.tx import Deserializer, DeserializerSegWit, DeserializerAuxPow, \
     DeserializerZcash, DeserializerTxTime, DeserializerReddcoin
 from server.block_processor import BlockProcessor
-from server.daemon import Daemon, DashDaemon, LegacyRPCDaemon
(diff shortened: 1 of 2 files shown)


bauerj:
Thanks that looks a lot better! Do you know why FujiCoin has no concept of dynamic minimum fees (or why it doesn't provide the RPC calls)?

fujicoin:
The release of Fujicoin is June 2014.
I think that the version of BTC-core of the forked coin is old.

fujicoin:
In coins using the old btc-core, if "Use dynamic fees" is used in electrum, the fee will not return from electrum-X, so the user can not remit money.
This is a problem also occurring in monacoin.
Other old coins are probably the same.
I think that there is the idea that this problem should be handled by the application side including electrum.
What shall we do?

kyuupichan:
This is fine.  Can you please add a comment to that effect to the implementation code, so anyone reading it understands why it is there.  Then I'll merge.

fujicoin:
Hi kyuupichan,

I add explanation of FujiDaemon.

Thanks,
motty

PR #226, opened 2017-08-20
Outgoing connections should use the same IP we're listening on

mmouse-:
This is important on a machine with more than one IP if a specific HOST= is configured so we're not listening on all IPs.  
Avoids source-destination mismatches when advertizing our peer.

### DIFF
--- a/server/peers.py
+++ b/server/peers.py
@@ -560,9 +560,11 @@ class PeerManager(util.LoggedClass):
             create_connection = self.proxy.create_connection
         else:
             create_connection = self.loop.create_connection
+            
+        local_addr = (self.env.host, None) if self.env.host else None    
 
         protocol_factory = partial(PeerSession, peer, self, kind)
-        coro = create_connection(protocol_factory, peer.host, port, ssl=sslc)
+        coro = create_connection(protocol_factory, peer.host, port, ssl=sslc, local_addr=local_addr)
         callback = partial(self.connection_done, peer, port_pairs)
         self.ensure_future(coro, callback)
 


server/peers.py
```diff
@@ -560,9 +560,11 @@ def retry_peer(self, peer, port_pairs):
             create_connection = self.proxy.create_connection
         else:
             create_connection = self.loop.create_connection
+            
+        laddr=(self.env.host, None) if self.env.host else None
```
kyuupichan: Can you name this local_addr like the param and put spaces around the assignment operator on this line.  Then I'll merge.  Thanks.

kyuupichan: And please add a comment to explain the problem it's solving.

kyuupichan: Do you not need to set the port in the tuple?  I can't test as I don't have this setup.

mmouse-: > Do you not need to set the port in the tuple? I can't test as I don't have this setup.  

AFAIK source port selection is normally done by the operating system, as each outgoing connection does need its own source port. So with many connections it would be very problematic to choose a port beforehand which is not occupied.  
Python docs state that you can use None to pass null to system layer: https://docs.python.org/3/library/socket.html#socket.getaddrinfo  

That said, it works for me in both cases (with and without HOST= set). But it should certainly receive some testing, especially in other (non-Linux) environments.

Regarding the other changes, I'm no programmer and my Python knowledge is minimal. But I'll do my best ;)

kyuupichan: Ah of course, it was a silly question.

PR #233, opened 2017-08-25
Various improvements

erasmospunk:
Code cleanups.

Replaced FujiDaemon with FakeEstimateFeeDaemon that simulates estimate fee calls and provide the same functionality.

Removed the parameter False for LegacyRPCDaemon's getblock RPC call as it is not needed.


kyuupichan:
Thanks

PR #243, opened 2017-09-01
add scripthash methods and bump protocol version

ecdsa:
this adds RPCs required for native segwit addresses

### DIFF
--- a/server/controller.py
+++ b/server/controller.py
@@ -97,9 +97,14 @@ class Controller(util.LoggedClass):
         # Set up the ElectrumX request handlers
         rpcs = [
             ('blockchain',
-             'address.get_balance address.get_history address.get_mempool '
-             'address.get_proof address.listunspent '
+             'address.get_balance '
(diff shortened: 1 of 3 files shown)


kyuupichan:
This is only part of what needs to be done:

- protocol version negotiation, only accept the RPC methods valid for the chosen protocol version
- version range lives in version.py
- all "address" RPC methods (excepting obsolete ones) to be provided for scripthash too
- implementation of other protocol changes for 1.1 (as documented)
- doc updates

Nevertheless the calls you added will be in the final version, so if it's useful for testing please use it.  The protocol negotiation will probably need a client-side change, as will the other protocol changes I mention above.

ecdsa:
May I suggest to create a release with only these changes, and to postpone your other improvements for another version number? The client is almost ready for a segwit release, and servers need to be upgraded a few days in advance.

fresheneesz:
Is this all that was needed for electrum server segwit support?

Issue #251, opened 2017-09-07
peer protocol version is not returned by server.peers.subscribe


ecdsa:
server.peers.subscribe seems to return version "1.0" for all peers, regardless of their actual protocol version. I do not know where that "1.0" number comes from.

ecdsa:
```
diff --git a/lib/peer.py b/lib/peer.py
index 83d6334..28230a2 100644
--- a/lib/peer.py
+++ b/lib/peer.py
@@ -90,7 +90,7 @@ class Peer(object):
         '''Convert a version string, such as "1.2", to a (major_version,
         minor_version) pair.
         '''
-        if isinstance(vstr, str) and VERSION_REGEX.match(vstr):
+        if isinstance(vstr, str) and cls.VERSION_REGEX.match(vstr):
             if '.' not in vstr:
                 vstr += '.0'
         else:
@@ -247,12 +247,12 @@ class Peer(object):
     @cachedproperty
     def protocol_min(self):
         '''Minimum protocol version as a string, e.g., 1.0'''
-        return self._version_string('protcol_min')
+        return self._version_string('protocol_min')
 
     @cachedproperty
     def protocol_max(self):
         '''Maximum protocol version as a string, e.g., 1.1'''
-        return self._version_string('protcol_max')
+        return self._version_string('protocol_max')
 
     def to_tuple(self):
         '''The tuple ((ip, host, details) expected in response

```

PR #252, opened 2017-09-07
applied patch by ecdsa

luggs-co:
issue #251
fixed name typos and getting string from correct location


kyuupichan:
Thanks.  I did this in a different way given I recently wrote library functions for protocol string handling.

PR #262, opened 2017-09-14
fix bandwidth issue with 2.8.3 clients and chunk downloads

danny91:
Issue with 2.8.3 clients using lots of bandwidth as you can see this one using 12GB. Changing this to include 2.8.3 fixes the issue.

https://i.imgur.com/QdpIo9B.png

### DIFF
--- a/server/session.py
+++ b/server/session.py
@@ -252,7 +252,7 @@ class ElectrumX(SessionBase):
 
         index: the chunk index'''
         index = self.controller.non_negative_integer(index)
-        if self.client_version < (2, 8, 3):
+        if self.client_version <= (2, 8, 3):
             self.chunk_indices.append(index)
             self.chunk_indices = self.chunk_indices[-5:]
             # -2 allows backing up a single chunk but no more.


kyuupichan:
This line seems to be in permanent incremental creep.  Perhaps it's time to admit the issue simply isn't fixed?  Can you point to something in 2.8.3 not in 2.8.2 that fixes it?

danny91:
@kyuupichan Not sure what you mean other than once i have added the "=" my bandwidth usage has returned to normal. This on ONLY on a cash server though. I have not needed the "=" on the bitcoinsegwit chain.
I think the issue steams from users using 2.8.3 to connect to cash servers whether intentionally or not. Maybe as some sort of an attack or maybe by accident. Either way this seems to have solved the issue.

danny91:
@kyuupichan Still having the same issue even on the most recent electrumx. Without doing this manually i have connections using Gigabytes of upload.
ellectrumx

kyuupichan:
Ok can you just remove the condition so that stepping backwards more than 1 or 2 is always rejected. I was told it was fixed in Electrum but apparently not

kyuupichan:
This feature has been removed so I'm closing this.  get_chunk is not restricted any more.

Issue #267, opened 2017-09-19
Allow running as root


lukechilds:
There are obvious reasons why running as root would not normally be a good idea. However there are certain circumstances when it's perfectly ok. For example applications inside docker containers usually run as root because they should have full permissions inside their container. This is completely safe.

Having to create new users and groups unnecessarily complicates the Dockerfile and can cause permission issues when mounting volumes.

Would you be open to the idea of allowing root but just showing a warning? Or alternatively keeping the current exit on root behaviour but allow enabling root it with `ALLOW_ROOT=1` or similar?

If you're ok with this I'm happy to put a PR together. Great work btw!

PR #268, opened 2017-09-19
Add ALLOW_ROOT option

lukechilds:
Resolves https://github.com/kyuupichan/electrumx/issues/267

Default behaviour is unchanged. However when running as root the error message will now be:

```
RuntimeError: RUNNING AS ROOT IS STRONGLY DISCOURAGED!
You shoud create an unprivileged user account and use that.
If you'd like to continue as root, please run this again with ALLOW_ROOT=1
CRITICAL:root:ElectrumX server terminated abnormally
```

Running as root with `ALLOW_ROOT=1` will run without error.

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -64,6 +64,10 @@ Miscellaneous
 
 These environment variables are optional:
 
+* **ALLOW_ROOT**
+
+  Set this environment variable to "1" to allow running ElectrumX as root.
+
 * **NET**
 
   Must be a *NET* from one of the **Coin** classes in `lib/coins.py`_.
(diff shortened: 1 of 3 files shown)


kyuupichan:
Thanks.  Like other booleans I prefer to refer to non-empty rather than 1 so I updated it to that effect.

lukechilds:
@kyuupichan Great!

Any chance you could do a new release with this change? I'm using the release tags on this repo to version my Docker builds.

kyuupichan:
Not immediately; I'm travelling at the moment.

lukechilds:
Ah, no worries.

If you wouldn't mind commenting here when it's done so I get a notification that would be sweet 👍

lukechilds:
Any plans for a new version tag?

lukechilds:
@kyuupichan Any plans for a new version tag?

PR #271, opened 2017-09-20
fix DASH set_protocol_handlers

cipig:
fix NameError: name 'masternode_announce_broadcast_1_0' is not defined


kyuupichan:
Thank you

PR #273, opened 2017-09-27
Update ENVIRONMENT.rst: how to create empty var

bithobbes:
It's rather unpleasant that it defaults to True with this ~~bug~~ odd behavior of some text editors.


kyuupichan:
What bug?

I'm not certain this file is the place to describe basic Unix operations.

kyuupichan:
I'll make this unnecessary by stripping environment variables before converting them to booleans.

Issue #282, opened 2017-10-06
Bitcoin Testnet peers mixed up


SomberNight:
There are several servers listed under BitcoinCashTestnet and none for BitcoinSegwitTestnet.
I believe at least some but possibly all of these are actually BitcoinSegwitTestnet servers.
https://github.com/kyuupichan/electrumx/blob/05d6351686bfa34738ab2beedef43a049ef33337/lib/coins.py#L392-L409

See also:
https://github.com/spesmilo/electrum/blob/5127dabb0df823529161a392490ec0357ad8affb/lib/servers_testnet.json

PR #283, opened 2017-10-06
update peers for BitcoinSegwitTestnet and BitcoinCashTestnet

SomberNight:
See #282.

I've tested all servers for BTC testnet and BCH testnet listed in this current repo, so
https://github.com/kyuupichan/electrumx/blob/05d6351686bfa34738ab2beedef43a049ef33337/lib/coins.py#L395-L402
and also the BTC testnet servers from the Electrum repo
https://github.com/spesmilo/electrum/blob/5127dabb0df823529161a392490ec0357ad8affb/lib/servers_testnet.json
and also the BCH testnet servers from the Electron Cash repo
https://github.com/fyookball/electrum/blob/bbe2ef61c07e0bb386c0772982fbaaf97e8b3b03/lib/network.py#L72-L75

These are the ones that were working during my tests.

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -393,12 +393,7 @@ class BitcoinCashTestnet(BitcoinTestnetMixin, Coin):
     '''Bitcoin Testnet for Bitcoin Cash daemons.'''
     NAME = "BitcoinCash"
     PEERS = [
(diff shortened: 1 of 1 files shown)


kyuupichan:
Thanks.

PR #284, opened 2017-10-06
Add support for Emercoin

snvakula:


kyuupichan:
This needs tests for all new codepaths (see tests for other coins).  Why do you need to do URLs differently?

snvakula:
Emercoin uses https for RPC by default. But it can be http as well. So this is a reason to add such URLs.
All required tests will be added soon, by updating the pool request.

lib/coins.py
```diff
@@ -367,6 +368,54 @@ class BitcoinSegwit(BitcoinMixin, Coin):
...
+    @classmethod
+    def sanitize_url(cls, url):
+        # Remove surrounding ws and trailing /s
+        url = url.strip().rstrip('/')
+        match = cls.RPC_URL_REGEX.match(url)
+        if not match:
+            raise CoinError('invalid daemon URL: "{}"'.format(url))
+        if match.groups()[1] is None:
+            url += ':{:d}'.format(cls.RPC_PORT)
+        if not url.startswith('http://') and not url.startswith('https://'):
```
kyuupichan: I don't think we need a separate function just for this change.  I don't think it would hurt to be in the base class implementation.  Can you do that, then I'll apply this.  Thanks.

kyuupichan:
You can see from the coverage report that your new read_header isn't tested.  Can you add a test for that please?

snvakula:
Am I correctly understood you want to remove support for https?

kyuupichan:
No I'm saying your change should be in the base class for the URL.  It won't do any harm there.  No need to have a specialization in your class.

snvakula:
Ok, I see. I'll update during this week only.

Issue #290, opened 2017-10-16
Synced btc database for v1.2


fr3aker:
I'm not sure this is a good place to post this at all but since a fair bunch of people will have to resync when upgrading to v1.2 I created a torrent magnet link for a synced btc leveldb (up to block 490153).

    magnet:?xt=urn:btih:caa804f48a319b061be3884ac011656c27121a6f&dn=electrumx_1.2_btc_leveldb_490153

Not sure if this is worth adding to the docs or somewhere.

If you don't want this around please close the issue.

Note: since electrum can validate any transaction through SPV this should not cause any trust issues with an externally acquired database as far as I can think.

kyuupichan:
I'd accept a PR to add it to the docs.  Thanks.

PR #293, opened 2017-10-17
add info about downloading pre-synced database

fr3aker:
implements #290 

If I should add it to a different file instead let me know.


kyuupichan:
This is great, can you just make clear it's the segwit bitcoin mainnet chain?  There is BCC now and also the various testnets.

fr3aker:
updated

PR #300, opened 2017-10-29
use estimatesmartfee for fee estimation

SomberNight:
It would be nice to start using `estimatesmartfee` instead of `estimatefee`.

`estimatefee` is deprecated in Bitcoin Core on master (https://github.com/bitcoin/bitcoin/pull/11031), which might already cause problems to some server ops (if they forget to explicitly enable the RPC).
Also, users would probably appreciate the lower estimates.

Note however that the `estimatesmartfee` RPC has only been introduced around 0.14.2 (??), so this won't work with older daemons.

The implementation here only affects BitcoinSegwit and BitcoinSegwitTestnet, I could only test with these coins/nets; but this could be extended. There might be nicer ways of doing this, not sure; open to ideas.

Related: https://github.com/spesmilo/electrum/issues/2420

### DIFF
--- a/server/daemon.py
+++ b/server/daemon.py
@@ -20,6 +20,7 @@ import aiohttp
 
 from lib.util import LoggedClass, int_to_varint, hex_to_bytes
 from lib.hash import hex_str_to_hash
+from lib.jsonrpc import JSONRPC
 
 
 class DaemonError(Exception):
@@ -30,8 +31,7 @@ class Daemon(LoggedClass):
     '''Handles connections to a daemon at the given URL.'''
 
     WARMING_UP = -28
-
-    HTTP_NOT_FOUND = 404
+    RPC_MISC_ERROR = -1
 
     class DaemonWarmingUpError(Exception):
         '''Raised when the daemon returns an error in its results.'''
@@ -100,7 +100,7 @@ class Daemon(LoggedClass):
                     # If bitcoind can't find a tx, for some reason
                     # it returns 500 but fills out the JSON.
                     # Should still return 200 IMO.
-                    if resp.status in (200, 500):
+                    if resp.status in (200, 404, 500):
                         return await resp.json()
                     return (resp.status, resp.reason)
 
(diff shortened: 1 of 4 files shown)


kyuupichan:
Better to test if supported and remember the result, and send the appropriate RPC thereon.

SomberNight:
Ok, will look into that. Probably tomorrow.

phungvietnam:
up

SomberNight:
I've pushed a new commit. Can rebase and squash optionally.

What's your take on this implementation (or at least approach)?

Note that we could clear `self._available_rpcs` when the daemon comes online after it was down as it might have been replaced with a different daemon. I think this would need locking though, with e.g. a dedicated lock for `self._available_rpcs`.

server/daemon.py
```diff
@@ -194,6 +197,24 @@ def processor(result):
             return await self._send(payload, processor)
         return []
 
+    async def _is_rpc_available(self, method):
+        available = self._available_rpcs.get(method, None)
+        if available is None:
+            payload = {'method': method, 'id': self.next_req_id()}
+            data = json.dumps(payload)
+            result = await self._send_data(data)
+            if not isinstance(result, tuple):
```
kyuupichan: The general approach is what I had in mind.  I don't like the duplication here though.  Why not just call self._send_single like self.estimatefee does, and catch the thrown exception when not available?

SomberNight: Well as far as I can see there is no exception to catch. If we call `self._send_single`, that calls `self._send`, and we reach L138. Log shows: `ERROR:Daemon:HTTP error code 404: Not Found  Retrying occasionally...`

kyuupichan: I see.  It's a shame bitcoind's JSON implementation is so bad, this isn't the only example.  It should be returning a JSON not HTTP error.  Perhaps the core loop should translate that error code to a DeamonError exception.

I'll take a closer look when I'm back.

kyuupichan: I can't test for several days myself but perhaps adding 400 to the list of error codes at line 100 for resp.status will have the effect of throwing the error.  Can you try that?

kyuupichan: It seems like bitcoind uses HTTP error codes independently of the JSON, and sends the JSON too.

SomberNight: Looks like this will work. Will have something soon.

kyuupichan:
Looks good, thanks.

PR #297, opened 2017-10-22
Add Hush

leto:
This adds Hush to electrumX. This commit has been powering transactions on main-net via the superNET fork:

https://github.com/SuperNETorg/electrumx/commit/8e55df9a7b5ea9bac01019e9183e36cd45893e30

which they are already running in production.

Additionally, this code was used to do a cross-chain atomic swap between HUSH and KMD, some links:

alicepayment 0.1001 KMD https://kmd.explorer.supernet.org/tx/4e45b45d501fb5892e1409c181db8b4d63c060cb954bb4e4e06dafdb4461ca50
bobpayment 0.11299695 HUSH https://explorer.myhush.org/tx/cda7db0094d8ee0181e2b48ac5873de2d0586e021bbf69f39f9fa61b7f624e00
bobdeposit 0.12710906 HUSH https://explorer.myhush.org/tx/0d46062aa3a07c56ebfd7fc458a76f5f188dee1e925aa505c879dba68dfde48d

Thanks for Electrum, it's awesome open source software! 😸

PR #309, opened 2017-11-12
Add Hush support, part deux

leto:
This is an updated version of https://github.com/kyuupichan/electrumx/pull/297, which I rebased from the latest master to get the Emercoin changes, and then added some tests for normal and P2SH Hush addresses. Please let me know if you would like other types of tests :metal:


ingamx:
Tests passing . Nice leto

kyuupichan:
The coverage report for lib/coins.py does not show the 2 functions in Hush class being tested at all; that is my main concern.  

https://coveralls.io/builds/14158493/source?filename=lib%2Fcoins.py

leto:
@kyuupichan after submitting my PR, I found the test data files for each chain, and I added one for Hush. At first I thought it was just the output of `hush-cli getblock` but then I realized it was the output of the ElectrumX RPC, since it contains the `block` header that our normal JSON does not contain, i.e. the deserialized data. I am currently trying to get valid test JSON data, and will update this PR when I have that working.

FYI, the 2 functions I added are exactly the same as the Zcash functions, Hush did not change any of that. I was thinking of maybe making a Mixin, but didn't want to change too much. Do you have thoughts on that?

leto:
@kyuupichan I have been digging deep, but I do not know how to generate the `block` key in the test data for each coin listed here: https://github.com/kyuupichan/electrumx/tree/master/tests/blocks

I have almost-correct test data here: https://github.com/leto/electrumx/blob/hush/tests/blocks/hush_mainnet_420.json but my `block` data in invalid and causes test failures.

I don't see any ElectrumX RPC method that returns the exact JSON in the test data files, can you throw me a 🍖 ? If I need to hand-carve it, any documentation about that would be appreciated.

kyuupichan:
Testing the header function is easy - just pass it the raw 140 binary bytes of a header and assert the result is what you expect.
Similarly the other function takes the binary bytes of a block - you have one in your test already but it's a hex string rather than binary bytes - and returns the header as binary bytes.  That header can be passed into the first function.  That way you're testing both at once.

ingamx:
how sweet 👍

erasmospunk:
@leto please use the new `EquihashMixin` to remove duplicate code (be sure to `git rebase` to the latest master)

Check commit https://github.com/kyuupichan/electrumx/commit/f3fea9f8ee88305aa087bb996a3d1857708b3a20 that added support to `EquihashMixin`.

erasmospunk:
@leto Also regarding the test json file, just use the `getblockhash` and `getblock` to get the decomposed block from the daemon. Remove any extra fields that you don't see in other coins and call `getblock` with an additional `false` argument to get the raw block and add it to the `"block"` field.

leto:
This was redone in https://github.com/kyuupichan/electrumx/pull/319, closing this

PR #314, opened 2017-11-16
Added support for sibcoin, changed decode message in jsonrpc

serbernar:


lib/jsonrpc.py
```diff
@@ -438,12 +438,16 @@ def decode_message(self, payload):
 
         Messages that cannot be decoded are logged and dropped.
         '''
-        try:
-            payload = payload.decode()
-        except UnicodeDecodeError as e:
-            msg = 'cannot decode message: {}'.format(e)
-            self.send_error(msg, JSONRPC.PARSE_ERROR)
-            return
+        encodings = ["utf-8", "iso-8859-1", 'cp-1251']
```
fr3aker: JSON is defined as allowing only unicode encoding (RFC 4627). Adding additional server overhead to deal with faulty implementations seems wasteful.

serbernar: I've tested this on electrum-dash-old client (which works on python2) but electrumx can't parse json, so for me it was a solution of a problem

kyuupichan:
The change to lib/coins.py is fine.  I won't change the JSON code.  Please remove those changes and I will merge.

serbernar:
Yep, i does it

PR #316, opened 2017-11-17
Clear peers for sibcoin

serbernar:


serbernar:
I forgot to change peers, sorry

PR #319, opened 2017-11-26
Add Hush using the shiny new EquihashMixin

leto:
This adds the Zcash fork Hush in a very maintainable way, using the EquihashMixin and then using the Zcash deserializer so that no new code is introduced *and* it comes with test data grabbed from a Hush 1.0.9 full node. All block tests passed locally for me on Python 2.7.13, please let me know if you need anything else. 

This PR replaces https://github.com/kyuupichan/electrumx/pull/309


kyuupichan:
Thank you

PR #322, opened 2017-11-28
add REVS, JUMBLR, WLC and COQUI

cipig:


kyuupichan:
Are you able to add a test for the electrum_header function of the Equihash mixin?

cipig:
I didn't add the EquihashMixin, it was already there and i assumed this is already covered by tests. I am unsure if i can write one, never done that before and i am not familiar with python.

leto:
@kyuupichan do you want a test that makes sure that we inherit from the EquihashMixin correctly? That mixin already existed, as @cipig said, and so we are not sure of the kind of test you want. If you describe a bit more about what kind of tests you would like, I would love to write them 😺

kyuupichan:
You've already written a test for a block.  Just pass that block to the electrumx_header function (as binary bytes, not a hex string) and check the return value is correct.  I assume it works now so just test it returns what it does now.  That is all.

PR #324, opened 2017-12-04
Update HOWTO.rst

mpdas108:
Add explanation that you cannot connect to your electrumx server until it has finished synchronizing once.


kyuupichan:
Thanks.  I've pushed something a little different.

mpdas108:
Thanks.

PR #330, opened 2017-12-14
New protocol with mempool fees RPC

ecdsa:
- mempool.get_fees returns a histogram of mempool transactions per fee rate
- bump protocol version to 1.2

### DIFF
--- a/server/controller.py
+++ b/server/controller.py
@@ -825,6 +825,14 @@ class Controller(ServerBase):
         number = self.non_negative_integer(number)
         return await self.daemon_request('estimatefee', [number])
 
+    def mempool_get_fee_histogram(self):
+        '''Memory pool fee histogram.
+
+        TODO: The server should detect and discount transactions that
+        never get mined when they should.
+        '''
+        return self.mempool.get_fee_histogram()
+
     async def relayfee(self):
         '''The minimum fee a low-priority tx must pay in order to be accepted
         to the daemon's memory pool.'''
--- a/docs/PROTOCOL.rst
+++ b/docs/PROTOCOL.rst
@@ -774,6 +774,20 @@ Subscribe to a script hash.
     [**scripthash**, **status**]
 
 
+mempool.get_fee_histogram
+=========================
+
+Return a histogram of the fee rates paid by transactions in the memory
+pool, weighted by transaction size.
+
+The histogram is an array of (fee, vsize) values, where vsize_n is the
+cumulative virtual size of mempool transactions with a fee rate in the
+interval [fee_(n-1), fee_n)], and fee_(n-1) > fee_n.
+
+Fee intervals may have variable size. The choice of appropriate
+intervals is currently not part of the protocol.
+
+
 server.add_peer
 ===============
 
--- a/server/mempool.py
+++ b/server/mempool.py
@@ -25,7 +25,7 @@ class MemPool(util.LoggedClass):
 
     To that end we maintain the following maps:
 
-       tx_hash -> (txin_pairs, txout_pairs)
+       tx_hash -> (txin_pairs, txout_pairs, tx_fee, tx_size)
        hashX   -> set of all tx hashes in which the hashX appears
 
     A pair is a (hashX, value) tuple.  tx hashes are hex strings.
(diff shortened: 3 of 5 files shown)


kyuupichan:
Thanks Thomas.

I'm not keen on this adding more fields to the Tx datastructure.  That is intended to be a simple representation of a Tx, and keeping it minimal in code and data size is important for initial sync.  Moreover this feature is only a need of a single chain ElectrumX supports but all chains are paying the cost.

Since this functionality is about mempools only, I would like this to be isolated to mempool code that processes Txs.  Mempool processing is fast on most chains as mempools are small, so that seems a better place for the burden.

If tx_hash -> (txin_pairs, txout_pairs, tx_fee, tx_size) is the best data structure (I'm not sure; I've not thought about it much), I think we should make that a named tuple.

It seems you're recomputing the histogram every time there is no mempool processing to do.  This might be a bit frequent and CPU intensive, particularly with 150k txs in the mempool.  What do you think of every 10 minutes and every new block?  I'm not sure how you are implementing this on the client side - is it queried every time a user tries to construct a tx to send, or is it just regularly polled?

The new API should only be provided for 1.2 clients (see how other RPC methods are restricted by client protocol version).

Finally is @fireduck64 on board with adding this to jelectrum?

kyuupichan:
Oh and please update docs/PROTOCOL.rst.  I note the 100,000 number seems to be a hard-coded part of this protocol.  I'm not sure that's appropriate; I think that should be returned by the protocol response.  At some future time the server could be changed to adjust so that the histogram has some appropriate granularity and the protocol wouldn't have to change.

server/controller.py
```diff
@@ -825,6 +825,14 @@ def block_get_header(self, height):
         number = self.non_negative_integer(number)
         return await self.daemon_request('estimatefee', [number])
 
+    async def mempool_get_fees(self):
```
kyuupichan: This doesn't need to be async as it isn't (in the current implementation at least; perhaps it should be if it potentially involved a lazy histogram recompute).  The JSON infrastructure in ElectrumX does not require callbacks to be async - see server.features and block.get_chunk.

fireduck64:
I'm game for adding this to jelectrum.  However, I still haven't fixed my addresses for the script hash change.

ecdsa:
I do not think the 100,000 parameter should be returned in the response, because that implies that the currently implemented constant-size intervals are part of the protocol. That should not be the case. For example, it might be interesting to use fee intervals that increases exponentially when the fee rate goes down, because less resolution is needed for the bottom of the memory pool. I want to leave that open for now.

ecdsa:
mempool.get_fee will be polled by the client every minute, and not only to contruct a new transaction. That is because mempool information will be displayed in the history tab for every unconfirmed transaction.

I have added a time interval of 30 seconds between recomputations of the histogram. I am not sure about the best way to trigger a recomputation on each new block; the code mentions a race condition in on_new_block, but I guess we could set an event from there.

ecdsa:
Regarding the addition of an integer field to the Tx data structure: I do not think that it would significantly impact performance, given that this data structure already contains arrays. 

To avoid this, we could return the transaction size in deserializer.read_tx, but that would affect more lines of code than the current proposal. 

Finally, completely isolating the size computation in mempool.py would probably require to parse every transaction twice, and that would be far worse for performance than the addition of a simple field in Tx.

ecdsa:
Latest commit uses exponentially increasing intervals, as discussed above.

ecdsa:
@kyuupichan any reason for not merging this?

kyuupichan:
@ecdsa - yes it's penalizing all other coins for the issues faced by one.  I don't want the Tx structure to bloat, and I don't like the implementation of the auto-compute of the histogram.  That should be done lazily (and recomputed if e.g. it's > 30s since the prior compute).  Bitcoin Cash clients, for example, have no forseeable reason to implement requesting the fee histogram, so it doesn't make sense the server is calculating stuff frequently that will likely never be requested.

I can try and modify your code to work this way, but it could take a week or two.

kyuupichan:
@ecdsa I've changed the read_tx() interface of lib/tx.py.  I noticed only 1 caller needed the hash, so I've split out returning both into a separate function.  Also created a new read_tx_and_vsize() call.  That is what I'd like your patch to use from the mempool code.  It also means the TX object doesn't need to add extraneous fields.  Does that seem reasonable?

What do you think about renaming your RPC method to mempool.get_fee_histogram?  Seems better to me.

ecdsa:
All right, it seems reasonable.
I rebased this PR against current master. 
It uses read_tx_and_vsize(), and histogram is computed lazily

kyuupichan:
I want to make changes but I'll do those in my own follow-up.  Can you confirm you've tested this from the Electrum client and it works as you intend?

ecdsa:
yes it works as intended.
it has been tested with the mempoolfees branch of the electrum repository.

PR #333, opened 2017-12-15
Link to newer Docker image

### LINKED ISSUE #331: Link to newer Docker image in the readme
I created my own Docker image ([`lukechilds/electrumx`](https://github.com/lukechilds/docker-electrumx)) due to some issues I had with the current image recommended in the readme ([`followtheart/electrumx`](https://github.com/followtheart/electrumx-docker)).

The followtheart image doesn't follow Docker best practices and is also quite large and out of date. I believe my image is better because:

**Size**

`lukechilds/electrumx`: 40MB
`followtheart/electrumx`: 175MB

**Follows Docker conventions**

- All settings are passed in as ENV vars (no need for ENV folder)
- Runs ElectrumX directly as PID 1 instead of bash -> daemon
  - Interactive Docker sessions attach directly meaning `Control-C` works
  - Interactive Docker sessions get ElectrumX output to STDOUT
  - ElectrumX output will be piped to normal Docker logging tools
  - You shouldn't dockerize a daemon, the Docker image itself should be daemonized
- Reproducible builds: separate branch for each release (followtheart container builds current master)

**Up to date**

Pulling from Docker Hub will pull the latest ElectrumX release. You can also specify a tag to pull specific older ElectrumX releases.

The latest followtheart build on Docker Hub is 6 months out of date.

**Secure**

If no SSL cert is provided one will be automatically generated on first boot.

---

Following these rules makes the image much easier to orchestrate in other Docker projects (e.g [Dockerized Vertcoin Electrum Server](https://github.com/lukechilds/docker-electrum-vertcoin/blob/master/docker-compose.yml)), it's also currently the most used ElectrumX image on Docker Hub.

Let me know if you'd like to link to this image instead and I'll happily submit a PR.

lukechilds:
Closes #331

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,7 +22,7 @@ There is also an `Dockerfile`_ available .
 
 .. _installer: https://github.com/bauerj/electrumx-installer
 
-.. _Dockerfile: https://github.com/followtheart/electrumx-docker
+.. _Dockerfile: https://github.com/lukechilds/docker-electrumx
 
 Features
 ========


lukechilds:
@bauerj No problem, good questions.

> 1. I have seen both mounting config files and env vars to configure the image. Is there actually a standard document that prefers either one?

The "twelve-factor app" methodology specifically states config should be set as environment variables: https://12factor.net/config

Also, Docker Compose allows you to compose containers together with environment variables. Allowing this project to be implemented in a single file and deployed with a single command: https://github.com/lukechilds/docker-electrum-vertcoin/blob/master/docker-compose.yml

That would be much messier with config files.

> 2. Apparently you generate both the certificate and the key if either of them is missing. Wouldn't it make more sense to do that only if both of them are missing?

Good spot, yes it would, I'll fix that.

> 3. How do you make sure that new releases will get a corresponding image? Is there some sort of automation on your side or do you check regularly for new releases?

Building, testing and publishing is handled automatically by CI, however I need to manually create a new `vX.X.X` branch and update the `VERSION` file to trigger this.

If any collaborators can ping me after a new release, that's great. If not, I'll check back regularly.

> 4. Can you add support for rocksdb too? (Maybe a separate flavour if you are concerned about image size)

I seem to remember having problems getting it working on Alpine Linux. Would happily accept a PR for this though. I could release it under a separate tag, like `lukechilds/electrumx:rocksdb`.

PR #337, opened 2018-01-01
Add Zclassic

heyrhett:
This add Zclassic similar to the same way that Hush was added here:
https://github.com/kyuupichan/electrumx/pull/319


kyuupichan:
Are you sure this is passing the test you added?

heyrhett:
Sorry, I needed to fix some things.  I confirmed it is passing the json test now.

PR #339, opened 2018-01-02
Workaround for Travis build failures

bauerj:
Apparently something in the release of pytest 3.3.0 broke our (arguably quite hacky) database closing method:

https://github.com/kyuupichan/electrumx/blob/master/tests/server/test_storage.py#L27-L29

This reverts to pytest 3.2 for now until I find out how to fix this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
   - "nightly"
 # command to install dependencies
 install:
+  - pip install pytest==3.2.5
   - pip install aiohttp
   - pip install ecdsa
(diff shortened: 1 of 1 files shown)


kyuupichan:
Hey Johann, thanks for investigating.  You made me realise my pytest was old and upgrading I could repro.  The fix is actually really easy and I've committed it - just call db.close() rather than doing the GC thing - that only applies to RocksDB, and is in its close method.

So no need for this any more.  Thanks again!

PR #341, opened 2018-01-03
Update: install_electrumx.sh

duckartes:
Remove "raspi-copies-and-fills" as it breaks the upgrade process


kyuupichan:
I have no way of testing this - I don't use Linux - so I'll take your word for it

PR #350, opened 2018-01-17
Added HOdlcoin support.

AmiT177:


AmiT177:
Travis compiler is missing plyvel ^^
code compiles perfectly.

kyuupichan:
This is fine, but can you please add a block test like most coins now have?  It means your coin is less likely to break when things change.

AmiT177:
Added

PR #354, opened 2018-01-24
Add NYC support

xarakas:


kyuupichan:
Please remove the IRC stuff; it is no longer supported.  Also please add a block test like exists for most other coins.

kyuupichan:
The test isn't passing

xarakas:
Fixed, thanks!

PR #358, opened 2018-01-29
Add Koto

wo01:


wo01:
I modified the source code to reuse Zcash deserialiser.

PR #389, opened 2018-02-26
Add custom base58check algorithm

erasmospunk:
There are coins like Decred that use a different base58check algorthms and this commit allows this kind of customisations.

### DIFF
--- a/lib/coins.py
+++ b/lib/coins.py
@@ -70,6 +70,8 @@ class Coin(object):
     BLOCK_PROCESSOR = BlockProcessor
     XPUB_VERBYTES = bytes('????', 'utf-8')
     XPRV_VERBYTES = bytes('????', 'utf-8')
+    ENCODE_CHECK = Base58.encode_check
+    DECODE_CHECK = Base58.decode_check
     # Peer discovery
     PEER_DEFAULT_PORTS = {'t': '50001', 's': '50002'}
     PEERS = []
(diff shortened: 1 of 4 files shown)


lib/hash.py
```diff
@@ -143,18 +143,18 @@ def encode(be_bytes):
         return txt[::-1]
 
     @staticmethod
-    def decode_check(txt):
+    def decode_check(txt, hash_fn=double_sha256):
```
kyuupichan: For this and encode_check please insert a *:

def decode_check(txt, *, hash_fn=double_sha256)

lib/coins.py
```diff
@@ -168,7 +169,8 @@ def address_to_hashX(cls, address):
     def P2PKH_address_from_hash160(cls, hash160):
         '''Return a P2PKH address given a public key.'''
         assert len(hash160) == 20
-        return Base58.encode_check(cls.P2PKH_VERBYTE + hash160)
+        return Base58.encode_check(cls.P2PKH_VERBYTE + hash160,
+                                   cls.BASE58_CHECK_HASH)
```
kyuupichan: And this to be cls.ENCODE_CHECK(cls.P2PKH_VERBYTE + hash160)

lib/coins.py
```diff
@@ -70,6 +70,7 @@ class Coin(object):
     BLOCK_PROCESSOR = BlockProcessor
     XPUB_VERBYTES = bytes('????', 'utf-8')
     XPRV_VERBYTES = bytes('????', 'utf-8')
+    BASE58_CHECK_HASH = double_sha256
```
kyuupichan: I think I'd prefer ENCODE_CHECK=Base58.encode_check.

Then decred can use a partial.  What do you think?

However a way to avoid unconditionally importing the new hash lib needs to be found....

erasmospunk:
@kyuupichan updated

kyuupichan:
You star, thanks a lot.

PR #396, opened 2018-03-05
Add Denarius (DNR)

carsenk:
Adding Denarius (DNR), Tests are in /tests folder.


setup.py
```diff
@@ -9,7 +9,7 @@
     python_requires='>=3.6',
     # via environment variables, in which case I've tested with 15.0.4
     # "x11_hash" package (1.4) is required to sync DASH network.
-    install_requires=['plyvel', 'pylru', 'aiohttp >= 1'],
+    install_requires=['plyvel', 'pylru', 'tribus_hash', 'aiohttp >= 1'],
```
kyuupichan: Please leave this out.  Users will have to install this manually if they want it.  Note there is no x11_hash, or scrypt specified there either.

kyuupichan:
Also the test is failing in travis

carsenk:
Seems to be failing due to missing tribus_hash 
pip install tribus_hash

Any work around? Pushing removal of it again.

carsenk:
So build still fails with tribus_hash removed because it seems it can't match the test block, which would be fair to assume....reverted commenting out of tribus_hash, x11_hash is called in the same manner, but does not error out on build in Travis?

kyuupichan:
The package for pip is tribus-hash.  Not sure how x11 works.  But add the install command to .travis.yml to get tribus-hash.  Then it should pass right?

carsenk:
Passed build. :)

PR #397, opened 2018-03-05
Enable verbose mode in blockchain.transaction.get

gdassori:
This PR enable a limited subset of the verbose mode from the getrawtransaction API.
The inclusion blockhash of a given transaction can be now obtained, along with the raw transaction hex.

### DIFF
--- a/docs/PROTOCOL.rst
+++ b/docs/PROTOCOL.rst
@@ -934,11 +934,7 @@ Protocol version 1.2 introduces new methods `blockchain.block.headers`,
   and support will be removed in some future protocol version.  You should update your code to use `blockchain.block.headers`
   and `Script Hashes`_ with the scripthash methods introduced in protocol 1.1 instead.
 
-* `blockchain.transaction.get` now have the optional parameter *verbose*. In verbose mode (default=False) the
-  result is enriched with the relative inclusion blockhash, if any.
-
-
-
+* `blockchain.transaction.get` now has an optional parameter *verbose*.
 
 blockchain.block.headers
 ========================
@@ -1014,17 +1010,18 @@ blockchain.transaction.get
 
 Return a raw transaction.
 
-  blockchain.transaction.get(**tx_hash**, **verbose** = [False])
+  blockchain.transaction.get(**tx_hash**, **verbose**=False)
 
   **tx_hash**
 
     The transaction hash as a hexadecimal string.
 
-  **verbose** [optional]
-    Set verbose mode (default = False).
+  **verbose**
+    Verbose mode, passed on to bitcoind in **getrawtransaction**.
 
 **Response**
 
    In non-verbose mode return the raw transaction as a hexadecimal string.
 
-   With verbose=True, return a JSON object with the deserialized transaction.
+   If verbose, returns what your coin's bitcoind returns; see its
+   documentation for details.
--- a/docs/PROTOCOL.rst
+++ b/docs/PROTOCOL.rst
@@ -532,7 +532,7 @@ blockchain.transaction.get
 
 Return a raw transaction.
 
-  blockchain.transaction.get(**tx_hash**, **height**)
+  blockchain.transaction.get(**tx_hash**, **height**, **verbose** = [False])
 
   **tx_hash**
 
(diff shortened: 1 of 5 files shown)


docs/PROTOCOL.rst
```diff
@@ -532,7 +532,7 @@ blockchain.transaction.get
 
 Return a raw transaction.
 
-  blockchain.transaction.get(**tx_hash**, **height**)
+  blockchain.transaction.get(**tx_hash**, **height**, **verbose** = [False])
```
kyuupichan: No, this 1.0 API with height isn't changing.

You need to make a copy under "Version 1.2" and remove the height parameter (see the notes under Version 1.1).   I know the layout of the protocol information is awkward; I will clean that up later.

docs/PROTOCOL.rst
```diff
@@ -924,17 +954,18 @@ Get a list of features and services supported by the server.
       "hash_function": "sha256"
   }
 
-Protocol Version 1.2
---------------------
-
```
kyuupichan: Please do not remove this - the structure of the file is currently by version.  At some stage I will switch it to be by method, but that's a big change...

server/daemon.py
```diff
@@ -262,9 +262,9 @@ def processor(result):
         network_info = await self.getnetworkinfo()
         return network_info['relayfee']
 
-    async def getrawtransaction(self, hex_hash):
+    async def getrawtransaction(self, hex_hash, verbose=False):
         '''Return the serialized raw transaction with the given hash.'''
-        return await self._send_single('getrawtransaction', (hex_hash, 0))
+        return await self._send_single('getrawtransaction', (hex_hash, verbose))
```
kyuupichan: we want int(verbose) because as the Bitcoin Core docs indicate, the integer form is more backwards-compatible than the boolean

return await self._send_single('getrawtransaction', (hex_hash, int(verbose)))

server/controller.py
```diff
@@ -842,13 +842,26 @@ def mempool_get_fee_histogram(self):
...
+        if verbose and isinstance(transaction, dict):
+            try:
+                return {
+                    'hex': transaction['hex'],
+                    'blockhash': transaction.get('blockhash')
+                }
+            except KeyError:
+                self.logger.exception('verbose mode failed')
+                raise RPCError("error in verbose mode")
+        return transaction
```
kyuupichan: As I indicated yesterday let's simply return what the network daemon returns and keep it simple.  Just change line 851 to 

return await self.daemon_request('getrawtransaction', tx_hash, verbose)

so it's only a 2-line change - that line and the async def line.

The Daemon class infrastructure will automatically return a good error if an error occurs; I designed things so being a user of the class is simple.

You will probably need to update your tests.

docs/PROTOCOL.rst
```diff
@@ -924,17 +954,18 @@ Get a list of features and services supported by the server.
...
- `blockchain.address.` are deprecated and support will be removed in
- some future protocol version.  You should update your code to use
- `blockchain.block.headers` and `Script Hashes`_ with the scripthash
- methods introduced in protocol 1.1 instead.
+* `blockchain.block.get_chunk` and all methods beginning `blockchain.address.` are deprecated
+  and support will be removed in some future protocol version.  You should update your code to use `blockchain.block.headers`
+  and `Script Hashes`_ with the scripthash methods introduced in protocol 1.1 instead.
+
+* `blockchain.transaction.get` now have the optional parameter *verbose*. In verbose mode (default=False) the
+  result is enriched with the relative inclusion blockhash, if any. Feature may have be unavailable.
```
kyuupichan: Please describe this as simply returning what the network daemon returns; we will not make any guarantees about what that might be.

gdassori: Ok :-)

server/controller.py
gdassori: Looks like tests are ok, since the response is mocked and the behaviour is almost the same.

server/daemon.py
gdassori: Mhm, I so have to better understand the code duplication here, I thought this method was invoked by the one in the Controller, and so the verbose param is casted as int in the previous class and passed here always as an int.
However, it's not an issue to do again it, I guess.

kyuupichan:
Thanks for the good tests

PR #400, opened 2018-03-06
Add coin SNG

blackjok3rtt:
This coin has been added. I have connected, received balance and sent one coin without issue. Can you please merge to main repo. Thanks.


lib/coins.py
```diff
@@ -883,6 +883,22 @@ class Zcash(EquihashMixin, Coin):
     RPC_PORT = 8232
     REORG_LIMIT = 800
 
+class snowgem(EquihashMixin, Coin):
```
kyuupichan: Please call the class and its NAME SnowGem following the convention of other coin names.  That is also what snowgem.org does.  The rest is good.

blackjok3rtt: Done, thanks for fast reply. :)

PR #409, opened 2018-03-11
Changes for BTG

wilsonmeier:


kyuupichan:
@erasmospunk does this look good to you, as the original submitter?

wilsonmeier:
Hi @kyuupichan

sorry for not adding more information with the pull request :)
I am a member of the BitcoinGold Team (see https://github.com/BTCGPU) and we are currently implementing an electrum wallet. 

This pull request extends the current electrum header to include the equihash solution. We need that to be able to validate the solution on client side. 
It adds params for our upcoming testnet v3, which introduces a new diff algorithm, as well as node regtest mode for easier development.
You can look up the new testnet params at https://github.com/h4x3rotab/BTCGPU/blob/genesis/src/chainparams.cpp

Let me know if you need anything more.

wilsonmeier:
Thx :D

PR #410, opened 2018-03-12
bitbayd rawblock newer daemon

GiorgosK:
bitbayd has adapted new method of creating raw blocks


GiorgosK:
Not sure what this coverage means or how to correct it

kyuupichan:
Thanks!

PR #424, opened 2018-03-23
add AXE support (new class)

charlesrocket:


kyuupichan:
You Daemon and Session classes seem identical to Dash's.  Why create new ones?

charlesrocket:
@kyuupichan wanted to have some space to build on top in the future. should i convert with (Dash)? was about to add x11

kyuupichan:
Well it seems nothing is new.  So either just reuse their classes, or if you want to keep your own, inherit theirs and add nothing.  Are you likely to want to add something that they won't have?  It seems your coin is a Dash clone.

charlesrocket:
@kyuupichan you right, this is almost same code at this time, exact same for electrumx classes. i am cloning classes for future development, so we could pull this repo and build right on top of it. but looks like i need to add x11 for tests as well. so maybe it would be better to convert this PR to Dash classes. what would you suggest?

charlesrocket:
noticed weird crash during wheel build on nightly. just tested on https://travis-ci.org/AXErunners/axe_hash could travis glitch? they went to maintenance after these commits

erasmospunk:
This needs rebase and squashing everything in one commit.

charlesrocket:
@erasmospunk guess it was a bad idea to go with one basket. any chance i could fix it with write access? 😞

kyuupichan:
Not sure what's happened but your PR is a mess now (e.g. it's touching 13 files and has conflicts).  Can you fix and resubmit?

charlesrocket:
@kyuupichan could you set tmp write access so i could have a look to fix it since there is another commits besides mines @erasmospunk requested complete squash, but i don't have write access to finish rebase and fix conflicts.

kyuupichan:
It's all at your end, write access is not needed

charlesrocket:
@kyuupichan you okay with resubmitting? can i go with dedicated branch to avoid same overlapping?

kyuupichan:
Of course

charlesrocket:
@kyuupichan got it, will resubmit

PR #429, opened 2018-03-29
add AXE support

charlesrocket:
new class


kyuupichan:
Are you going the new class route?  

Can you:

1) Please put 2 blank lines before class declarations, not 1
2) inherit the Dash classes and do not redefine functions that are the same.

If 2) means the classes just have a body of "pass" that is fine, but you should expect that if they remain the same for an extended time I will merge them at some point.

Please update with those changes.

charlesrocket:
i woud like to keep new classes if its ok, will merge by your request later if needed

charlesrocket:
working on 3.7 resolution, don't want to change travis file

charlesrocket:
can not fix the issue with p3.7 at this time, will wait until official release. so moved classes

PR #432, opened 2018-03-30
Add DROP_CLIENT env variable

erasmospunk:
This will disconnect any client that matches a regular expression.
This is useful for dropping buggy/unsupported clients.

### DIFF
--- a/docs/environment.rst
+++ b/docs/environment.rst
@@ -201,6 +201,12 @@ These environment variables are optional:
 
   If you are not sure what this means leave it unset.
 
+.. envvar:: DROP_CLIENT
+
+  Set a regular expression to disconnect any client based on their
+  version string. For example to drop versions from 1.0 to 1.2 use
+  the regex ``1\.[0-2]\.\d+``.
+
 
 Resource Usage Limits
 =====================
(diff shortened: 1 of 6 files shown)


server/env.py
```diff
@@ -67,6 +67,8 @@ def __init__(self):
         self.max_session_subs = self.integer('MAX_SESSION_SUBS', 50000)
         self.bandwidth_limit = self.integer('BANDWIDTH_LIMIT', 2000000)
         self.session_timeout = self.integer('SESSION_TIMEOUT', 600)
+        self.drop_client = self.custom("DROP_CLIENT", None,
+                                       lambda v: re.compile(v))
```
kyuupichan: No lambda, just re.compile

docs/environment.rst
```diff
@@ -201,6 +201,11 @@ These environment variables are optional:
 
   If you are not sure what this means leave it unset.
 
+* **DROP_CLIENT**
+
+  Disconnect any client that matches this regular expression. This is useful
+  for disconnecting buggy clients.
+
```
kyuupichan: Please update this to match the surrounding doc format (you're using the old style).  Also be more specific about what exactly is being matched, and perhaps give an example value.

erasmospunk:
Thanks for the notes, updated.

PR #451, opened 2018-04-13
fix comment

wakiyamap:


kyuupichan:
Thanks, I updated it to point to readthedocs instead.

PR #457, opened 2018-04-25
coins.py update for Viacoin

marshallstokes:
Update list of electrumx servers for Viacoin
Truncate auxpow from headers


kyuupichan:
Can we have some tests for Viacoin please; see the block tests for other coins.  Ideally there would be one for your overridden block_header function.

I don't see much benefit in lots of hard-coded servers for one domain.

kyuupichan:
Any updates?  Otherwise this will be closed.

PR #463, opened 2018-05-05
Add Uniform Fiscal Object (UFO)

Bushstar:
I set the following name...

`NAME = "UniformFiscalObject"`

Seemed to be inline with other names and saves having to wrap the name in quotes in the conf file. ElectrumX is now running with UFO and fully synced on five servers.


kyuupichan:
Your test is failing it appears

PR #469, opened 2018-05-15
Added ZCL testnet, BTCP mainnet, and BTCP testnet

csulmone:


kyuupichan:
Please add block tests for the coins.

csulmone:
Will add these, sorry for the delay

PR #474, opened 2018-05-23
Add $PAC

elmora-do:
- $PAC support added.
- $PAC masternode tracking feature added.
- $PAC blocks for tests added.


server/daemon.py
```diff
@@ -309,6 +309,16 @@ class DashDaemon(Daemon):
         '''Return the masternode status.'''
         return await self._send_single('masternodelist', params)
 
+class PacDaemon(Daemon):
```
erasmospunk: Please use DashDaemon instead of copy-pasting.

server/session.py
```diff
@@ -587,3 +588,144 @@ def notify(self, height, touched):
             self.mns.add(vin)
             return result.get(vin)
         return None
+
+class PacElectrumX(ElectrumX):
```
erasmospunk: Please extend the DashElectrumX and override any methods. Even better, extend DashElectrumX if your additions apply to Dash other master-node coins as well.

elmora-do:
I refactored the code in order to:
- Use DashDaemon instead of PacDaemon.
- Extend DashElectrumX and override the electrumx handlers to add 2 more endpoints.

server/session.py
```diff
@@ -553,10 +553,15 @@ def notify(self, height, touched):
...
-                                   [masternode, status.get(masternode)])
+        try:
+            for masternode in self.mns:
+                status = self.daemon.masternode_list(['status', masternode])
+                self.send_notification('masternode.subscribe',
+                                       [masternode, status.get(masternode)])
+        except DaemonError as e:
+            error, = e.args
+            message = error['message']
+            self.logger.info('there was an error when notifying the masternode status: {}'.format(message))
```
elmora-do: I added this try/catch block because sometimes I got this error:

```
electrumx_server.py[13104]: ERROR:Controller:Traceback (most recent call last):
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/controller.py", line 178, in on_future_done
electrumx_server.py[13104]:     future.result()
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/block_processor.py", line 206, in main_loop
electrumx_server.py[13104]:     await task()
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/block_processor.py", line 256, in check_and_advance_blocks
electrumx_server.py[13104]:     self.controller.mempool.on_new_block(self.touched)
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/mempool.py", line 194, in on_new_block
electrumx_server.py[13104]:     self.controller.notify_sessions(touched)
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/controller.py", line 296, in notify_sessions
electrumx_server.py[13104]:     session_touched = session.notify(height, touched)
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/session.py", line 608, in notify
electrumx_server.py[13104]:     'result': status.get(masternode),
```
I know it's not the best way but it's just a workaround, I'll try to fix it later.

server/session.py
```diff
@@ -552,10 +553,16 @@ def notify(self, height, touched):
...
+        try:
+            for masternode in self.mns:
+                status = self.daemon.masternode_list(['status', masternode])
+                if hasattr(status, 'get'):
+                    self.send_notification('masternode.subscribe',
+                                           [masternode, status.get(masternode)])
+        except DaemonError as e:
+            error, = e.args
+            message = error['message']
+            self.logger.info('there was an error when notifying the masternode status: {}'.format(message))
```
elmora-do: I added this try/catch block because sometimes I got this error:

```
electrumx_server.py[13104]: ERROR:Controller:Traceback (most recent call last):
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/controller.py", line 178, in on_future_done
electrumx_server.py[13104]:     future.result()
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/block_processor.py", line 206, in main_loop
electrumx_server.py[13104]:     await task()
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/block_processor.py", line 256, in check_and_advance_blocks
electrumx_server.py[13104]:     self.controller.mempool.on_new_block(self.touched)
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/mempool.py", line 194, in on_new_block
electrumx_server.py[13104]:     self.controller.notify_sessions(touched)
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/controller.py", line 296, in notify_sessions
electrumx_server.py[13104]:     session_touched = session.notify(height, touched)
electrumx_server.py[13104]:   File "/usr/local/lib/python3.6/dist-packages/electrumx-1.3.1-py3.6.egg/server/session.py", line 608, in notify
electrumx_server.py[13104]:     'result': status.get(masternode),
electrumx_server.py[13104]: AttributeError: 'coroutine' object has no attribute 'get'
```
I know it's not the best way but it's just a workaround, I'll try to fix it later.

server/session.py
```diff
@@ -142,7 +143,7 @@ def __init__(self, *args, **kwargs):
         self.hashX_subs = {}
         self.mempool_statuses = {}
         self.protocol_version = None
-        self.set_protocol_handlers((1, 0))
+        self.set_protocol_handlers((1, 3))
```
kyuupichan: This is not wanted

server/session.py
```diff
@@ -587,3 +594,95 @@ def notify(self, height, touched):
             self.mns.add(vin)
             return result.get(vin)
         return None
+
+class PacElectrumX(DashElectrumX):
```
kyuupichan: Is there any reason this new code does not apply to Dash too?

server/session.py
kyuupichan: You're right, it's not a workaround at all.  You will get that every time.  The clue is in the error message - it's a coroutine because you missed the "await".

server/session.py
```diff
@@ -587,3 +594,95 @@ def notify(self, height, touched):
...
+        else:
+            result = self.controller.mn_cache
+        return result
+
+    async def masternode_info(self, address):
+        '''Returns the full info of masternode.'''
+        result = []
+        address_list = ''.join(address.split()).split(',')
+        result = await self.masternode_list(address_list)
+        return result
```
kyuupichan: Ideally there would be a test for this functionality

server/session.py
elmora-do: yeap, I tried by adding the 'await' but I got 'syntax error' because the 'notify' method is not async, also I tried using the asyncio library to wait until the thread is done but I got another error

server/session.py
elmora-do: there's no reason, that code can apply to Dash as well, I just didn't want to modify it (what if I broke something), but shall I add it to the DashElectrumX class? 

btw the code was tested with the PAC network and works well

server/session.py
elmora-do: oh okay, I've never create that kind of tests but I'll add some

server/session.py
kyuupichan: I see.  Check out the notify_async function - which I created for async notification things.  Ultimately this will all be async, but for now see if you can modify the way that works to achieve what you want.

server/session.py
kyuupichan: Yes please add it to DashElectrumX

server/session.py
kyuupichan: I won't hold merging if there are no tests because much of the other code here is also untested (for now).  But then if it breaks I won't be worried about fixing it either...

server/daemon.py
```diff
@@ -309,7 +309,6 @@ class DashDaemon(Daemon):
         '''Return the masternode status.'''
         return await self._send_single('masternodelist', params)
 
-
```
kyuupichan: Since there is no change to this file let's not remove the line

kyuupichan:
One more thing - the docs are nicely up-to-date other than for the masternode RPCs.  Can you please document the masternode RPCs in the same fashion as the other RPCs are documented?

server/session.py
elmora-do: question, at what moment notify_async is triggered? because I'm figuring out how it works

elmora-do: I already tested and that code can be implemented in `notify_async `or in `notify `either using the `controller.create_task `method.

That's why I'm asking about `notify_async `

kyuupichan:
This looks good.  Just one thing remains I think - updates to docs/protocol-methods, perhaps in a DASH section, that documents the dash-specific RPCs.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
   "ElectrumX 1.2.1"
+
+masternode.announce.broadcast
+--------------
+
+Pass through the masternode announce message to be broadcast by the daemon.
+
+**Signature**
+
+  .. function:: masternode.announce.broadcast(masternode_announce_message)
```
kyuupichan: The parameter names are important and form part of the RPC API, and for that reason the code and the documentation must match for parameter names.  This parameter name is too long and in the code it is "signmb".

How about just "message"?

I'm curious - why is Electrum Dash responsible for broadcasting masternode announcements?  Isn't this something the Dash daemon does itself?

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+  
+**Result**
```
kyuupichan: Each parameter requires an explanatory section - see other RPC docs.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+
+  A message with the result of the broadcasting, it could be a success message or an error one including the reason of failure.
```
kyuupichan: Results are never errors.  Just describe what is returned on success as this is intended to document the "result" item in the dictionary; any RPC can theoretically return an error.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+
+masternode.subscribe
+--------------
+
+Returns the status of masternode.
+
+**Signature**
+
+  .. function:: masternode.subscribe(masternode_collateral)
```
kyuupichan: Parameter must match the code.  How about "collateral" or just "address" (I'm guessing but I assume that's what it is).  Please document the parameter.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+  
+  * *payees*
+
+    A single string or an array of masternode payee addresses.
+
+**Result**
+
+  An array with the masternodes information.
+  
+  * If the payee is empty, all the masternodes in the network will be retrieved.
```
kyuupichan: I assume this means an empty array?  Or does it mean missing?  Or does it mean either?  Please clarify.

Please refer to the parameter as "payees" as that is what you named it, not "payee".

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+**Examples**::
+
+  masternode.list()
+  masternode.list("9d298c00dae8b491d6801f50cab2e0037852cb556c5619ddb07c50421x9a31ab")
+  masternode.list("['9d298c00dae8b491d6801f50cab2e0037852cb556c5619ddb07c50421x9a31ab',
+  '9d298c00dae8b491d6801f50cab2e0037852cb556c5619ddb07c50421x9a31ac']")
+
+**Example Results**::
+    
+  When nothing is sent all masternodes information is received.
```
kyuupichan: "Nothing"?

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+    ]
+
+masternode.info
+--------------
+
+Returns the full info of masternode.
+
+**Signature**
+
+  .. function:: masternode.info(masternode_payee)
```
kyuupichan: Perhaps "address" is sufficient?  Please make the name match the code.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,178 @@ Identify the client to the server and negotiate the protocol version.
...
+
+    Masternode payee address.
+
+**Result**
+
+  An array with the full information of the masternode.
+  
+**Examples**::
+
+  .. masternode.info("9d298c00dae8b491d6801f50cab2e0037852cb556c5619ddb07c50421x9a31ab")
```
kyuupichan: This doesn't look like a Dash address.  Is it a public key?  Or a TX ID?  The documentation of each method should be clear about all the inputs.

server/session.py
```diff
@@ -495,22 +496,29 @@ def set_protocol_handlers(self, ptuple):
...
         return result
 
+
     # Masternode command handlers
     async def masternode_announce_broadcast(self, signmnb):
         '''Pass through the masternode announce message to be broadcast
-        by the daemon.'''
+        by the daemon.
+
+        signmnb: masternode announce message.'''
```
kyuupichan: You are not doing any type checking.  It is being passed through to the daemon, so that may be acceptable if the daemon does the type checking.

server/session.py
```diff
@@ -521,9 +529,111 @@ def notify(self, height, touched):
...
+                mn_info['lastpaidblock'] = mn_data[6]
+                mn_info['ip'] = mn_data[7]
+                mn_info['paymentposition'] = get_payment_position(mn_payment_queue, mn_info['payee'])
+                balance = await self.controller.address_get_balance(mn_info['payee'])
+                mn_info['balance'] = sum(balance.values()) / self.controller.coin.VALUE_PER_COIN
+                mn_list.append(mn_info)
+            self.controller.mn_cache = mn_list
+
+        if payees:
+            if type(payees) is str:
```
kyuupichan: Please use isinstance(payees, str) to be consistent with the rest of the codebase.

It seems here that an empty string has the same behaviour as an empty array, or the number 0, or "null".  I don't like the loosenesss here.  Can we tighten it to require an argument and for it to be a string or an array?  I would like to not have a default so that the caller must explicitly pass an argument.  That will require updating the function signature (and perhaps the docs).

server/session.py
```diff
@@ -521,9 +529,111 @@ def notify(self, height, touched):
...
+                result = [mn for mn in self.controller.mn_cache if mn['payee'] == payees]
+            else:
+                result = [mn for mn in self.controller.mn_cache for address in payees if mn['payee'] == address]
```
kyuupichan: Please wrap this line (and any others) that are over 79 bytes.  the 2nd line can start with the for or the if.

server/session.py
```diff
@@ -521,9 +529,111 @@ def notify(self, height, touched):
...
+        return result
+
+    async def masternode_info(self, address):
+        '''
+        Returns the full info of masternode.
+
+        address: masternode payee address.
+        '''
+        result = []
+        address_list = ''.join(address.split()).split(',')
```
kyuupichan: This will crash if address is not a string - you are using untrusted input in all these functions so all must be type checked for their operations.

docs/protocol-methods.rst
elmora-do: Ok I'm going to update the parameter names in order to match the docs and the code.

No, it doesn't, every time you create a masternode you have to create an announcement message and broadcast it to the network manually.

And thanks for the RPC explanation I didn't know that.

docs/protocol-methods.rst
elmora-do: sorry my bad, it should be an address.

server/session.py
elmora-do: yes, the daemon does all the validations

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,98 @@ Identify the client to the server and negotiate the protocol version.
...
+masternode.announce.broadcast
+--------------
+
+Pass through the masternode announce message to be broadcast by the daemon.
+
+Whenever a masternode comes online or a client is syncing, they will send this message which describes the masternode entry and how to validate messages from it.
+
+**Signature**
+
+  .. function:: masternode.announce.broadcast(signmnb)
```
kyuupichan: What does signmnb mean?  Is it signed?  If so, can you update the description to include the word signed?

If not why not just call the parameter "message"?

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,98 @@ Identify the client to the server and negotiate the protocol version.
...
+masternode.subscribe
+--------------
+
+Returns the status of masternode.
+
+**Signature**
+
+  .. function:: masternode.subscribe(vin)
+
+  * *vin*
```
kyuupichan: We can do better than "vin".  how about "address"?  What does "collateral" mean?  Reading the parameter name and its description I have no idea what should be passed.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,98 @@ Identify the client to the server and negotiate the protocol version.
...
+
+    Masternode collateral.
+
+**Result**
+
+  As this is a subcription, the client will receive a notification when the masternode status changes.
```
kyuupichan: What is the "status" of a masternode?

server/session.py
```diff
@@ -521,9 +528,102 @@ def notify(self, height, touched):
...
+            self.controller.mn_cache = mn_list
+
+        # If payees is an empty list the whole masternode list is returned
+        if payees:
+            result = [mn for mn in self.controller.mn_cache 
+            for address in payees if mn['payee'] == address]
+        else:
+            result = self.controller.mn_cache
+
+        result = sorted(result, key=lambda x: x['paymentposition'])
```
kyuupichan: Is sorting guaranteed?  If so please document it; otherwise I don't think we should sort.

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,106 @@ Identify the client to the server and negotiate the protocol version.
...
+
+Whenever a masternode comes online or a client is syncing, they will send this message which describes the masternode entry and how to validate messages from it.
+
+**Signature**
+
+  .. function:: masternode.announce.broadcast(signmnb)
+
+  * *signmnb*
+
+    Signed masternode broadcast message.
```
kyuupichan: It's still not clear to me what this is like and you don't give any examples.  Can you do that in a follow-up?

docs/protocol-methods.rst
```diff
@@ -1055,3 +1055,106 @@ Identify the client to the server and negotiate the protocol version.
...
+
+Returns the status of masternode.
+
+**Signature**
+
+  .. function:: masternode.subscribe(collateral)
+
+  * *collateral*
+
+    A masternode collateral is a transaction with a specific amount of coins, it's also known as a masternode identifier.
```
kyuupichan: So what is passed?  A transaction ID, or a transaction?  In hex?  How does it know which output is relevant?

Please add an example in a follow-up.

server/session.py
```diff
@@ -520,10 +527,103 @@ def notify(self, height, touched):
...
+                mn_info['ip'] = mn_data[7]
+                mn_info['paymentposition'] = get_payment_position(mn_payment_queue, mn_info['payee'])
+                mn_info['inselection'] = mn_info['paymentposition'] < mn_payment_count // 10
+                balance = await self.controller.address_get_balance(mn_info['payee'])
+                mn_info['balance'] = sum(balance.values()) / self.controller.coin.VALUE_PER_COIN
+                mn_list.append(mn_info)
+            self.controller.mn_cache = mn_list
+
+        # If payees is an empty list the whole masternode list is returned
+        if payees:
```
kyuupichan: I still see no typechecking.  If a string is passed it will do something weird here.  Please raise an error if it's not an array (in a follow-up).

kyuupichan:
Thanks, I've merged this.  Can you address my 3 comments above in a follow-up please.

elmora-do:
Sure, I'll do it

PR #482, opened 2018-05-31
Add minimum confirmations for getting balance

jesobreira:
I think it would be good if users were able to decide how many confirmations they need in order to declare a transaction actually confirmed (recommended is 6, Electrum uses 1).

This has been requested (spesmilo/electrum#3779), and also required changes on Electrum (spesmilo/electrum#4391).

Unfortunately I don't have enough storage for deploying and testing my own PR, so please consider this as an untested idea.


kyuupichan:
This is a presentation layer GUI thing, doesn't belong in the server

PR #490, opened 2018-06-08
update version aiorpcx

wakiyamap:


kyuupichan:
Weird; could have sworn I made that change locally.  Got lost somehow.  Thanks.

PR #495, opened 2018-06-17
Add TokenPay

gabriel-samfira:
This commit adds TokenPay as a supported coin. It also adds a new
function to TxInput, that returns True if the input is spendable.
An input can be marked as unspendable, if for example, it originates
from a stealth address.


server/block_processor.py
```diff
@@ -540,6 +540,12 @@ def advance_txs(self, txs):
             # Spend the inputs
             if not tx.is_coinbase:
                 for txin in tx.inputs:
+                    # NOTE(gabriel-samfira): Be forgiving if someone wants
+                    # to rebase and has not implemented is_spendable in their
+                    # custom TxInput
+                    check_spendable = getattr(txin, 'is_spendable', None)
+                    if check_spendable and check_spendable() is False:
+                        continue
```
kyuupichan: I'm sorry but having something coin-specific in these 2 core loops is not acceptable for performance reasons

server/db.py
```diff
@@ -243,7 +243,7 @@ def fs_block_hashes(self, height, count):
         headers_concat, headers_count = self.read_headers(height, count)
         if headers_count != count:
             raise self.DBError('only got {:,d} headers starting at {:,d}, not '
-                               '{:,d}'.format(headers_count, start, count))
+                               '{:,d}'.format(headers_count, height, count))
```
kyuupichan: Thanks for this; I'll push this separately

server/block_processor.py
gabriel-samfira: Okay, fair enough. Will look for alternatives. I must note, that this will probably happen with any coin that has stealth addresses, not just this one. So it may benefit other coins as well. Any suggestion as to where this would be better placed?

Thanks for the quick review :D.

gabriel-samfira:
Will do, thanks! :)

server/block_processor.py
kyuupichan: The problem with your approach is a) the test is inefficient, but more b) it is done per tx_in.

I would prefer some kind of pre- or post-processing loop that did whatever was necessary.

Also, I think you will need to do something for block reorgs, because that requires undoing what was done when processing a block, so tends to incorporate similar logic.  I don't see that in your code.

I would like to require tests for new functionality such as this, but as I don't (yet) have them for the standard code paths I feel it's hypocritical to ask others to do that; realistic tests for block processing will require a fair amount of work.

kyuupichan: I may have mis-spoken and you it seems you did cover the reorg case.

gabriel-samfira: Indeed. These changes were needed to have electrumx work with stealth transactions. Inputs originating from stealth inputs have [a different format](http://paste.ubuntu.com/p/jzYmYZD9kk/), and the ```scriptSig``` field includes [OP_RETURN](https://en.bitcoin.it/wiki/OP_RETURN).

Will think about a more efficient way to do this, but it is great to get input from you :). Apologies for not including tests as well.

gabriel-samfira: Apropos, a similar check to what I added, seems to happen in:

https://github.com/kyuupichan/electrumx/blob/master/server/block_processor.py#L545

```hashX_from_script()``` [seems to do the same kind of check](https://github.com/kyuupichan/electrumx/blob/master/lib/coins.py#L131-L137) and it is done for every output. The check is just done in another function.

I will still think about finding a better solution for my change though. Just thought it was worth mentioning :).

Comment (gabriel-samfira) [reply server/block_processor.py]: Hmm, I am not sure that within a transaction, you are unable to have a normal input and a stealth input being spent, just as you can have stealth outputs and non stealth outputs as shown in the pastebin I attached (line 89-101). In which case, we kind of have to go through each input.

**Comment by gabriel-samfira** [comment]
Sorry for the delay on this, will attempt to finish tests and push during the weekend.

File: server/block_processor.py (hunk 1 above)
Comment (gabriel-samfira) [reply server/block_processor.py]: Moved the is_spendable check in the coin class. For any coin without stealth transactions, it's a noop defaulting to ```True```.

Also made a drive-by fix on latest master. In ```getrawtransactions()```, the second parameter is supposed to be ```int``` not ```bool``` (as is the case in ```getrawtransaction()``` where it gets converted from bool to int before being put on the wire)

File: electrumx/server/daemon.py [hunk 3]
```diff
@@ -271,7 +271,7 @@ def processor(result):
         '''Return the serialized raw transactions with the given hashes.
 
         Replaces errors with None by default.'''
-        params_iterable = ((hex_hash, False) for hex_hash in hex_hashes)
+        params_iterable = ((hex_hash, 0) for hex_hash in hex_hashes)
```
kyuupichan: This needs its own PR.

electrumx/lib/coins.py
```diff
@@ -78,6 +78,13 @@ class Coin(object):
     PEER_DEFAULT_PORTS = {'t': '50001', 's': '50002'}
     PEERS = []
 
+    @classmethod
+    def txin_is_spendable(cls, tx_input):
+        """Some coins have implemented stealth transactions
+        In those cases, the script will be marked with OP_RETURN
+        and should be treated as not spendable."""
```
kyuupichan: The code already treats OP_RETURN as not spendable.  What makes you think otherwise?

electrumx/server/block_processor.py
```diff
@@ -534,6 +534,9 @@ def advance_txs(self, txs):
             # Spend the inputs
             if not tx.is_coinbase:
                 for txin in tx.inputs:
+                    # Verify that this is not a stealth input
+                    if self.coin.txin_is_spendable(txin) is False:
+                        continue
```
kyuupichan: Sorry adding coin-specific code to this key loop is not acceptable.  But I don't see why you need to customise this.

electrumx/lib/coins.py
gabriel-samfira: I feel like I have done a poor job at describing the initial issue. Apologies.

> The code already treats OP_RETURN as not spendable. What makes you think otherwise?


It does so for tx outputs (a check that is also called indirectly during the loop). For tx inputs it does not.

I changed the PR a bit to demonstrate the issue. Essentially, I removed the continue clause and added a log message including the script of the input. The diff compared to what is proposed here, is:

```diff
diff --git a/electrumx/server/block_processor.py b/electrumx/server/block_processor.py
index b895a9e..079c2ec 100644
--- a/electrumx/server/block_processor.py
+++ b/electrumx/server/block_processor.py
@@ -18,6 +18,7 @@ from functools import partial
 from electrumx.server.daemon import DaemonError
 from electrumx.lib.hash import hash_to_str, HASHX_LEN
 from electrumx.lib.util import chunks, formatted_time, class_logger
+from electrumx.lib.script import OpCodes
 import electrumx.server.db
 
 
@@ -535,8 +536,9 @@ class BlockProcessor(electrumx.server.db.DB):
             if not tx.is_coinbase:
                 for txin in tx.inputs:
                     # Verify that this is not a stealth input
-                    if self.coin.txin_is_spendable(txin) is False:
-                        continue
+                    if txin.script[0] == OpCodes.OP_RETURN:
+                        self.logger.warning("Found non spendable input: %r" % (txin.script,))
+                        #continue
                     cache_value = spend_utxo(txin.prev_hash, txin.prev_idx)
                     undo_info_append(cache_value)
                     append_hashX(cache_value[:-12])
@@ -617,8 +619,9 @@ class BlockProcessor(electrumx.server.db.DB):
             if not tx.is_coinbase:
                 for txin in reversed(tx.inputs):
                     # Verify that this is not a stealth input
-                    if self.coin.txin_is_spendable(txin) is False:
-                        continue
+                    if txin.script[0] == OpCodes.OP_RETURN:
+                        self.logger.warning("Found non spendable input: %r" % (txin.script,))
+                        #continue
                     n -= undo_entry_len
                     undo_item = undo_info[n:n + undo_entry_len]
                     put_utxo(txin.prev_hash + s_pack('<H', txin.prev_idx),
```

I removed the contents of the DB folder and ran the daemon again, forcing a sync. The result is as follows:

```log
$ electrumx_server 
WARNING:Env:lowered maximum sessions from 1,000 to 674 because your open file limit is 1,024
INFO:Controller:software version: ElectrumX 1.5
INFO:Controller:aiorpcX version: 0.5.6
INFO:Controller:supported protocol versions: 1.1-1.3
INFO:Controller:event loop policy: None
INFO:LegacyRPCDaemon:daemon #1 at 127.0.0.1:8800/ (current)
INFO:BlockProcessor:switching current directory to /home/electrum/electrum_db/
INFO:BlockProcessor:using leveldb for DB backend
INFO:BlockProcessor:created new database
INFO:BlockProcessor:creating metadata directory
INFO:History:flush count: 0
INFO:History:history DB version: 0
INFO:BlockProcessor:DB version: 6
INFO:BlockProcessor:coin: TokenPay
INFO:BlockProcessor:network: mainnet
INFO:BlockProcessor:height: -1
INFO:BlockProcessor:tip: 0000000000000000000000000000000000000000000000000000000000000000
INFO:BlockProcessor:tx count: 0
INFO:BlockProcessor:sync time so far: 00s
INFO:BlockProcessor:reorg limit is 500 blocks
INFO:BlockProcessor:flushing DB cache at 1,200 MB
INFO:Controller:RPC server listening on localhost:8000
INFO:Prefetcher:catching up to daemon height 241,815 (241,816 blocks behind)
INFO:Prefetcher:verified genesis block with hash 000008b71ab32e585a23f0de642dc113740144e94c0ece047751e9781f953ae9
INFO:BlockProcessor:our height: 9 daemon: 241,815 UTXOs 0MB hist 0MB
INFO:BlockProcessor:our height: 5,009 daemon: 241,815 UTXOs 2MB hist 2MB
INFO:BlockProcessor:our height: 10,009 daemon: 241,815 UTXOs 4MB hist 2MB
INFO:BlockProcessor:our height: 15,009 daemon: 241,815 UTXOs 6MB hist 3MB
INFO:BlockProcessor:our height: 20,009 daemon: 241,815 UTXOs 8MB hist 4MB
INFO:BlockProcessor:our height: 30,009 daemon: 241,815 UTXOs 13MB hist 5MB
INFO:BlockProcessor:our height: 37,509 daemon: 241,815 UTXOs 17MB hist 6MB
INFO:BlockProcessor:our height: 47,009 daemon: 241,815 UTXOs 22MB hist 9MB
INFO:BlockProcessor:our height: 54,509 daemon: 241,815 UTXOs 25MB hist 10MB
WARNING:BlockProcessor:Found non spendable input: b'j\xb9\xb0\x9bp\x0b\xea:N\xf3\xf7\xfe\xb9}\xcc!\x8d\xbcB\xcb\xb8\xc7w\x02\x13\xcc\xfd\x1f\xd5\xb1\x90\xc8\x8cug{\xf2\xb33\x0bmr\x97\xb8\xb3J\xd8\xb5\xb3d\x10\xad\x0e\xd3\xee2\x06\xdb\x8d8Z\x13\xf8>\xd8\xe0\x99!s^oY\x8a\xc88.c\x04\xbc\xec\x01 \xc7\x90\xb3\'\xd8\xc3\x94\x8e\x04b \xbbi\x1b3M\xban\xba?\xc0\x1e\xdb\xecK\x95\xd8}\xc52\xb0\x0b\xea\xbb\xf4$\x8c\xcc/O\x88\x84\xf1w\x85B\xb4\xe3G\xe4Z\xae\x9f\x19\xf2\x16#\xac1\xec\xbc\xc1\x8d\xba\x1e\xb6To\x1a\x07\xa4\xcaQ\x11qd\x19\xaf\xe7\x17B\x18F\xd3\xbfK\x1a03n\xb1\n\xa6&I\x1a\xef\x06`d\xa9-\x02\x9c\xc3?Sa\xbcu\xc1\x9aB\x948|\x8b\xd1\xbf"~\x04\xc2If\xfe\x81^\xb2$t\x16\np\x95\x9b\xb9\xfb\xeaX#v\x97\xd7\x02\xe4e\xa3\xdc\xb1o\x07\xcc\x93\xb3\xed\xda\xadh\xc2\xb5Z\x1cAk/\x19\xa1A\xed\n\tP\x01\xb7\xf5)\x03\xf2\xadX\x85\x83\x1d\xca\xd0\xd6I\'\x9a\xf6\x8e1o\x141f\xb0Y\x03^\xd9\xfd0\xc0G\x9a\x96\x92\x12\x02m\xdd\x1d\xf7OV^\xbc\x15b\xf3\xe9\x96\x86\x19O\xf5x\xafl<!=\xaa\x0fHD\xaeSc\x03\x1e\x02\xaa\x99Y\xb5E\x9d\x83\xcf\x00\xcdT\x0b\xb5\x18\xb4\x8bS\x9a\xac\xfe\xb3%\x8d\xd3\xb6h\xac_\x84\xe4\r\xcd\x03\xc9\xf6\x1c&68I)\xa3\x1a\x98\xd8C\xc3n\xa2\xc9e\xda\xae\x81e\xd3\xa0\x13IvT\xecf\xef\'\x03\xba\x0e\xa0\xa6e\x0eSR/\xad\xf4JHi\x11\xe4}\xe7\xfaZ\xa4\xdf\x04\x94\xbf\x02.\xb4\xcf\xc3\x92\xac'
ERROR:Controller:uncaught task exception: ushort format requires 0 <= number <= (0x7fff * 2 + 1)
Traceback (most recent call last):
  File "/home/electrum/electrumx/electrumx/server/controller.py", line 202, in check_task_exception
    task.result()
  File "/home/electrum/electrumx/electrumx/server/block_processor.py", line 203, in main_loop
    await task()
  File "/home/electrum/electrumx/electrumx/server/block_processor.py", line 247, in check_and_advance_blocks
    await self.controller.run_in_executor(self.advance_blocks, blocks)
  File "/home/electrum/electrumx/electrumx/server/controller.py", line 186, in run_in_executor
    return await self.loop.run_in_executor(None, func, *args)
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/electrum/electrumx/electrumx/server/block_processor.py", line 497, in advance_blocks
    undo_info = self.advance_txs(block.transactions)
  File "/home/electrum/electrumx/electrumx/server/block_processor.py", line 542, in advance_txs
    cache_value = spend_utxo(txin.prev_hash, txin.prev_idx)
  File "/home/electrum/electrumx/electrumx/server/block_processor.py", line 696, in spend_utxo
    idx_packed = pack('<H', tx_idx)
struct.error: ushort format requires 0 <= number <= (0x7fff * 2 + 1)
```

This happens for the reorg case as well.

kyuupichan: So your daemon is permitting invalid transactions?  Or these are valid and should be present but the inputs should be ignored?

electrumx/server/block_processor.py
kyuupichan: I'd like you to explain why an input is treated this way and what it means.  The input that you're ignoring and not allowing to be spent remains in the UTXO set?  Or was it never there in the first place?   In which case what do the prev tx_hash and prev_index mean?

electrumx/lib/coins.py
gabriel-samfira: These inputs are actually valid, but represent a stealth input. Please see:

https://en.bitcoin.it/wiki/OP_RETURN

under "OP_RETURN applications".

electrumx/server/block_processor.py
gabriel-samfira: So the input we are ignoring is actually a stealth input. Essentially, when someone sends a certain amount of coin to a stealth address, those coins get "burned". They get removed from the public ledger and added to the private one.

The reverse also happens when sending coins from a private address to a normal address. They essentially get minted in the public ledger. For these types of transactions, there is no ```prev_index``` to point to. Electrumx being a public index, I think it's safe to ignore these inputs, and only count any public outputs. This also happens in the reverse case. We ignore "invalid" (OP_RETURN) outputs, when a normal input sends coins to a stealth address.

Here is a high level (stratospheric) explanation of what happens:

https://github.com/tokenpay/tokenpay/blob/master/doc/Ring%20Signatures.md

gabriel-samfira:
I have not yet written the tests, because of the pending discussion regarding stealth transactions. Unfortunately, adding TokenPay depends on that bit also merging. Otherwise it will never fully sync or work. There are already stealth transactions on the chain.

If you feel this change is something that could merge, and give your approval for the stealth input case, I will happily write the tests and submit them.

kyuupichan:
I have no objection to merging TokenPay.  I don't want to add code to the busiest loop in the codebase that is performance-critical that only does something for a single coin, even a single "if" statement.

I think you could create a TX deserializer for your coin (see lib/tx.py) that drops these inputs from the tx.inputs array.  Would that work?  I don't see anything in the current code that requires it to be non-empty or that requires knowing the "real" index in the inputs array.  You could inherit the current deserializer and just override the one method to drop the inputs after the fact.

I think you would have the same problem in the mempool code too (I doubt you've seen that as you've not had it running at the tip of the chain when a new stealth TX comes in), so this seems cleanest.  However it might break the mempool code's transaction fee calculation.

The loop(s) in question are likely to change in the near future and I don't want to have to worry about breaking coin-specific code that I cannot test, either.

I don't mind a single coin-specific call in the functions in question (for example, a pre- or post- processing step) if that is preferable to you, but not in the loops.

gabriel-samfira:
Sounds fair!

I will try to do it with a TX deserializer. Thanks a lot for the input!

gabriel-samfira:
Made the changes you requested, and added a couple of tests. Not sure if the tests are enough. Looked at some of the other ones as an example. Does not seem to be more than address tests, block tests, and coin selection tests. If I missed anything, let me know.

kyuupichan:
Hi could you rebase?  Can you answer some questions:

- does this sync properly on TokenPay mainnet and maintain the mempool correctly?
- have you seen it handle a block reorg correctly?
- why do you need a different coinbase rule?  Do you really have stealth coinbases where the first input script begins with OP_RETURN?

Otherwise this looks good, thanks

gabriel-samfira:
Hi!

Yup, will rebase as soon as I get a chance.

1) It seems to sync and maintain mempool. It is currently used in the mobile wallet
2) yes, and it does so successfully, with the custom TX deserializer. Stealth transactions were the only "major" issue.
3) The coinbase rule is there because ```is_coinbase()``` seems to be called for [all transactions in a block](https://github.com/kyuupichan/electrumx/blob/master/electrumx/server/block_processor.py#L515) when advancing txs, as well as in the reorg case. If dealing with stealth transactions, we may only have stealth inputs, which all get filtered out, leaving ```vin``` with an empty array. Calling ```self.inputs[0]``` would in this case error out. So by default, while a stealth transaction will not be a coinbase transaction, a check still seems to be done, regardless. (and if I have not misread the code).

One request though, I would like to postpone merging this until after the weekend.

I've noticed something odd in the case where an address receives a stake reward for mining a block. The hashX for the output is different, and electrumx seems to loose track of history because of this. I suspect it's something to do with how the raw block gets deserialized, but did not have a chance to investigate yet. I have some time set aside this weekend for it, and I think this will be a fix that can be contained inside the coin logic.

I am aware that there is no way to stake on electrum, but some people may opt to import their desktop wallet seed inside the electrum wallet, in which case if I can fix this without bastardizing anything, I'd rather do it before merging.

History is properly recorded and kept track of for addresses that never stake, like the mobile wallet and electrum desktop wallet.

By the way, thanks for enforcing pep8 in tests :) (and for the amazing project). Would be awesome to see that in the desktop wallet code as well :D.

kyuupichan:
I'm not in a rush to merge so I'll wait until you're ready.  I think what you're trying here is a little edgy so am happy to wait for it to be stable.

I'm considering getting rid of is_coinbase but already coins are creeping in hacking it to mean other than what is says, such as Decred...  I might eliminate it and just special-case the first TX, and am wondering what is depending on is_coinbase catching anything other than just the 1st tx, always.

kyuupichan:
Oh, and if your special-casing of is_coinbase is not required and only "suspicious", please remove it, it is confusing.

gabriel-samfira:
Found the issue of the missing history. It seems that the block reward for mining a new block is being paid out using P2PK script. Electrumx uses the P2PKH hash to keep track of history in its DB.

I've created a custom ```_read_output()``` for the ```TokenPay``` class, that converts the script from P2PK to P2PKH before returning it. This way the rest of the server logic is left untouched.

I will clean it up during the weekend, and I will push it (still need to treat the case where the publick key is uncompressed - for correctness sake).

Regarding ```is_coinbase()```, I don't think I can remove it safely. It seems to get called in the main loop, and it will most likely err out with ```KeyError``` if I do. It would be great not to have to worry about coinbase transactions in the main loop, and maybe filter it out during a pre-processing phase of the block.

Thoughts?

kyuupichan:
There is no need to change anything to handle P2PK.  See the docs on script hashes; it has such examples.
is_coinbase is already gone.

gabriel-samfira:
Okay then, I need to check why the pubkey handler is not treating that case properly in the TPAY case. Using a modified bitcoin electrum wallet, I only see the P2PKH outputs in the history tab. Will investigate.

kyuupichan:
Yes - that's how Electrum works; you need to change Electrum not ElectrumX

kyuupichan:
Yeah there is quite a bit of dead code in coins.py; I will remove it.  A lot of it originates from when it was intended for a wallet implementation...  though some of the code you highlight was used in older versions of ElectrumX that indexed by address.  Now it indexes by scripthash, as that is independent of address format (e.g. BCH's cashaddress; also BTC's segwit lark).

We add to the DB not the hash of the PK but the hash of the script, as the function name indicates.  ElectrumX indexes by (hash of) the script, and Electrum subscribes to the (hashes of) P2PKH scripts only by default.  You just need to adjust Electrum to subscribe to P2PK scripts and ElectrumX will return the data and you would see it in Electrum.  Note that changing Electrum is harder than it sounds, as it only signs P2PKH inputs, so you'd need to extend its signature scheme too to be able to spend them.

kyuupichan:
https://github.com/kyuupichan/electrumx/commit/eb5c55107aa2da42365f2a93f6d8e689b92aabbe removes most of the dead code...

gabriel-samfira:
indeed. In that case, I will rebase the current branch during the weekend, and I think we can merge as is. Will hack the electrum wallet to subscribe to P2PK as well. It's a bit annoying, because I'll also have to join the 2 history results to show the user a coherent history, but that is outside the scope of ElectrumX.

As a sidenote, have you thought about decoupling the coins from ElectrumX, and using a classloader to load a particular coin? It would mean that you can only focus on the core app, not worry about coins at all, and each coin could be independently developed and packaged. So if you want to use a specific coin, it would be a matter of

```bash
pip install electrumx-tokenpay

electrum_server --coin electrumx.tokenpay.TokenPayClass
``` 

or just give the path to the class in the config :).

kyuupichan:
@erasmospunk has done a lot of work recently to support coins that spend fake inputs that look like coinbases.  This is for block processing and mempool handling.  His work may well have solved the issues you saw.  Please try testing against git HEAD.

gabriel-samfira:
Rebased. I tried a full sync using HEAD, and the standard ```DeserializerTxTime```, but ran into:

https://paste.ubuntu.com/p/kFtswpwjK4/

Works with the custom deserializer though.

kyuupichan:
I could merge this.   However I think a better approach (because I don't think yours gets the mempool cases right) is to resurrect @erasmospunk 's is_generation abstraction, because I think if your derived class just overrode that everything would work.  I'm also leery of simply dropping inputs from the array because it no longer really represents the transaction.

@erasmospunk what do you think?  If you're both on board I can restore that myself (or one of you could do it).

kyuupichan:
@gabriel-samfira but if you find the mempool code already works for you when these kinds of Txs are in the mempool, and you prefer to keep your approach, I'm happy to merge as-is.   Does the mempool work properly once serving clients?

erasmospunk:
The transaction `is_generation` method wasn't ideal because you could have mixed normal inputs with generation/coinbase ones.

@gabriel-samfira can you paste the hex of the failing transaction?

kyuupichan:
Right, I meant making it per input

gabriel-samfira:
Hi @erasmospunk,

Here is an example of a stealth transaction:

https://paste.ubuntu.com/p/jzYmYZD9kk/

This includes outputs to both stealth addresses and normal public addresses.

I do not feel religious about my approach. If you guys have a better suggestion on how to treat this, I am all for it.

erasmospunk:
@gabriel-samfira understood what happened here, you have reused the `COutPoint` as keyimage + ring size. The patches that Neil talks about were to support coins that were creating a transactions with outpoints same as in a generation transaction:
`0000000000000000000000000000000000000000000000000000000000000000ffffffff`

In your case the inputs appear like spends to non existing UTXOs.

So to make it work you need a custom deserialiser in any case. Now if you drop the inputs like you did originally or let Neil restore the `TxInput::is_generation` and set it in your deserialiser, is a matter of taste. Restoring `TxInput::is_generation` is more versatile though.

> Right, I meant making it per input

> @erasmospunk what do you think? If you're both on board I can restore that myself (or one of you could do it).

In that case it sounds good, do it only for the inputs though, the one in the transaction would be useless.

erasmospunk:
@gabriel-samfira as side note, I tried to compile your daemon on ubuntu 16.04 but fails in the configure phrase due to openssl errors. Also the codebase seems old, you should try moving to the modern Bitcoin base that has the `depends` and `config.site` for easier dependency tracking and deterministic builds.

kyuupichan:
Give me a couple of days, I'll get to restoring is_generation as is own commit.  Then I'll provide you a best guess update to tx.py for your coin which you'll need to test and maybe fix.  When it works just replace it in your PR as your own code and I'll merge it.

kyuupichan:
@gabriel-samfira sorry for the delay.  https://github.com/kyuupichan/electrumx/commit/ab17b4d529722a1588fcca4cf826a7056504fc4c adds is_generation.  You will need to subclass TxInput (if you haven't already) and provide a suitable override.  In your case it seems to be something that is also a property of the tx itself so you probably need to add a member to your inputs.

But then mempool and other stuff should just work for you without dropping inputs which I'm not keen on.

gabriel-samfira:
No worries! I will give this a test ASAP. Thanks for resurrecting this!

gabriel-samfira:
Made the change. Looks good! For the stealth inputs, I created a new ```TxInput``` type. We can't tell if an input is stealth or not until we read it, and look at the script, so I always deserialize an input to a standard input, check if it is an anonymous input, then reserialize it into a stealth ```TxInput```. This is extra work for not much benefit at this stage, but it is at least correctly representing the transaction input.

Let me know if you prefer to dump that, or change it in any way.

PR #505, opened 2018-06-30
Support BitcoinDiamond

Starry9:


lib/tx.py
```diff
@@ -503,3 +503,146 @@ def read_tx(self):
...
+
+        if present_block_hash is not None:
+            return TxBitcoinDiamondSegWit(version, present_block_hash, marker,
+                                          flag, inputs, outputs, witness,
+                                          locktime)
+        else:
+            return TxSegWit(version, marker, flag, inputs,
+                            outputs, witness, locktime)
+
+    def _read_tx_parts(self):
```
kyuupichan: This function seems to duplicate most of read_tx.

PR #507, opened 2018-07-03
fix for JSON value is not an integer as expected

cipig:
Revert a small change from https://github.com/kyuupichan/electrumx/commit/2cb89814b6c8df2d2e19f0a2192c2c7216964140 to fix error
`{"jsonrpc": "2.0", "error": {"code": 2, "message": "daemon error: {'code': -1, 'message': 'JSON value is not an integer as expected'}"}, "id": 1}` when doing blockchain.transaction.get


kyuupichan:
Sorry, just committed this independently.  Please fix your daemon to accept booleans like recent bitcoin and bitcoin cash code.

PR #511, opened 2018-07-09
electrumx package

eukreign:
This PR turns electrumx into a package so that it is easy to extend and use in other projects.

Examples of using the new package:
- https://github.com/lbryio/lbryumx/blob/63fab3edd791b9fb568f44e00eae632a6fcdc5a0/lbryumx_server.py#L20
- https://github.com/lbryio/orchstr8/blob/4d788213296bede00efae53b2d56af74d1171f0e/orchstr8/services.py#L90


electrumx/server/controller.py
```diff
@@ -101,6 +102,8 @@ def __init__(self, env):
         self.mempool = MemPool(self.bp, self)
         self.peer_mgr = PeerManager(env, self)
 
+        self.tcp_server_started = asyncio.Event()
+
```
kyuupichan: These 2 lines are extraneous; please remove

electrumx/server/controller.py
```diff
@@ -291,6 +294,7 @@ def close_servers(self, kinds):
             sslc = ssl.SSLContext(ssl.PROTOCOL_TLS)
             sslc.load_cert_chain(env.ssl_certfile, keyfile=env.ssl_keyfile)
             await self.start_server('SSL', host, env.ssl_port, ssl=sslc)
+        self.tcp_server_started.set()
```
kyuupichan: Likewise

electrumx/server/env.py
```diff
@@ -27,15 +27,18 @@ class Env(EnvBase):
     # Peer discovery
     PD_OFF, PD_SELF, PD_ON = range(3)
 
-    def __init__(self):
+    def __init__(self, coin=None):
```
kyuupichan: Please remove the changes to Env; let's just get the package restructuring done.

I don't see the need to change Env given you can just change the environment of the caller.

kyuupichan:
Another question - you've gone for absolute imports (electrum.xx.yy) inside the code itself.  Any reason to prefer that to relative imports?  I don't know the answer to that; I hope you can enlighten me.

eukreign:
Regarding relative vs. absolute imports, I can change it to relative imports if you prefer this. Just let me know.

Here is an explanation of this issue with a link to PEP 8 which recommends absolute imports:
https://stackoverflow.com/a/28154841

I personally always use absolute imports because it eliminates ambiguity and makes it really quick to skim the imports to see what's local vs. what's a 3rd party dependency (it's harder to notice the small dot at the beginning of local subpackage names). This is the same reason that PEP 8 recommends absolute imports:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)

But it also says relative imports are okay:

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

kyuupichan:
OK let's stick with absolute, thanks

eukreign:
I've moved the `contrib` directory back out of the package and dropped the commit that was extra fixes not completely related to the re-packaging. Let me know if there is anything else I need to do.

eukreign:
As far as what things are used from the electrumx package it's actually most things, see various files here for example:
https://github.com/lbryio/lbryumx/tree/master/lbryumx

One of the LBRY features is a block chain based "name" lookup system (claimtrie) so we have to overload most things in electrumx to make it work.

eukreign:
Once this is merged I can do a few more PRs to include the dropped commit and also for an improvement to the `logger` that makes it easier to configure.

kyuupichan:
I've merged this; let's see how it goes.

I don't recommend you start relying on the internals of ElectrumX as they are subject to large change (and are likely to change a lot - see roadmap).  It's not clear to me what you are trying to achieve - is there any explanation somewhere - but it might be worth considering if it's possible as a documented interface of ElectrumX that doesn't impose performance consequences if unused.

eukreign:
We have unit tests and have also built a generic bitcoin based currency integration testing library which we use for our own currency: https://github.com/lbryio/orchstr8

I'm not too worried about changes in electrumx breaking our stuff, if we miss something it's our fault for not having covered it in our tests.

Thank you for merging!

Some background about how LBRY works is here: https://lbry.tech/resources/lbry-claimtrie.html

PR #514, opened 2018-07-10
make it easier to configure logging for electrumx

eukreign:
This PR improves the logging support in electrumx making it easier to configure behavior across the entire package instead of having to target individual classes. For example:

```python
# DEBUG everything:
logging.getLogger('electrumx').setLevel(logging.DEBUG)

# just WARNINGs:
logging.getLogger('electrumx').setLevel(logging.WARNING)
```

### DIFF
--- a/electrumx/lib/env_base.py
+++ b/electrumx/lib/env_base.py
@@ -19,7 +19,7 @@ class EnvBase(object):
         pass
 
     def __init__(self):
-        self.logger = logging.getLogger(self.__class__.__name__)
+        self.logger = logging.getLogger(__name__).getChild(self.__class__.__name__)
         self.allow_root = self.boolean('ALLOW_ROOT', False)
         self.host = self.default('HOST', 'localhost')
         self.rpc_host = self.default('RPC_HOST', 'localhost')
--- a/electrumx/lib/env_base.py
+++ b/electrumx/lib/env_base.py
@@ -19,7 +19,8 @@ class EnvBase(object):
         pass
 
     def __init__(self):
-        self.logger = logging.getLogger(__name__).getChild(self.__class__.__name__)
+        self.logger = logging.getLogger(__name__)\
+            .getChild(self.__class__.__name__)
         self.allow_root = self.boolean('ALLOW_ROOT', False)
         self.host = self.default('HOST', 'localhost')
         self.rpc_host = self.default('RPC_HOST', 'localhost')
--- a/electrumx/lib/server_base.py
+++ b/electrumx/lib/server_base.py
@@ -36,7 +36,8 @@ class ServerBase(object):
         '''Save the environment, perform basic sanity checks, and set the
         event loop policy.
         '''
-        self.logger = logging.getLogger(__name__).getChild(self.__class__.__name__)
+        self.logger = logging.getLogger(__name__)\
+            .getChild(self.__class__.__name__)
         self.env = env
 
         # Sanity checks
--- a/electrumx/lib/server_base.py
+++ b/electrumx/lib/server_base.py
@@ -36,7 +36,7 @@ class ServerBase(object):
         '''Save the environment, perform basic sanity checks, and set the
         event loop policy.
         '''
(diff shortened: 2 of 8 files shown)


eukreign:
Python uses a hierarchical logging system. So, if you create a bunch of loggers like:
1. electrumx.server.SomeClass
2. electrumx.server.SomeOtherClass
3. electrumx.lib.AnotherClass

You can control logging at the root package level, at sub packages or at the class level:

1. `logging.getLogger('electrumx').setLevel(logging.DEBUG)` will set logging for everything under `electrumx.
1. `logging.getLogger('electrumx.server').setLevel(logging.DEBUG)` will set it for everything under `electrumx.server`.
1. `logging.getLogger('electrumx.server.BlockchainProcessor').setLevel(logging.DEBUG)` will set it for everything under `electrumx.server.BlockchainProcessor`.

The `__name__` is just the dot delimited module path. So, `logging.getLogger(__name__).getChild(self.__class__.__name__)` translates to something like `logging.getLogger('electrumx.server').getChild('BlockchainProcessor')`.

Also, see: https://docs.python.org/3/library/logging.html#logging.Logger.getChild

kyuupichan:
Thanks for the detailed info.

kyuupichan:
This is now more verbose in logging in an undesirable way.  Can that be fixed?
I'm getting:

INFO:electrumx.server.peers.PeerManager

in my logs; I'd rather just have the final component of the path

kyuupichan:
I cannot see in the docs how to change this line so that the prior behaviour is preserved:

https://github.com/kyuupichan/electrumx/blob/master/electrumx_server.py#L21

kyuupichan:
All right.  

If that is tricky, alternatively we create a package-level class called, um, Logging, that has a function to return the loggers that all the lines you changed call instead.  A class-level boolean variable indicates whether that function returns new-style or old-style loggers.  You would need to set that class-level variable in your client code.

eukreign:
I don't think it's going to be tricky to make a formatter that re-creates the old behavior on output.

From my experience the Python best practice is to just define a module level `log` variable like so:

```python
import logging
log = logging.getLogger(__name__)
```

Then just use `log` everywhere in that particular file / module. Class level loggers aren't as common, actually electrumx is the first project I've seen doing that. The problem with class level loggers is that if a subclass also uses this logger the output will deceptively provide the wrong module/class path.

```python
# in a.py
import logging
class Foo:
  def __init__(self):
    self.log = logging.getLogger(__name__).getChild(self.__class__.__name__)

# in b.py
class Bar(Foo):
  def do_something(self):
    self.log.info("info logged in b.Bar but will appear to come from a.Foo")
```

If you forgo class level loggers altogether it solves this confusion, the log messages will better indicate where they are from:

```python
# in a.py
import logging
log = logging.getLogger(__name__)
class Foo:
  def __init__(self):
    log.info("clear in output it's logged from a")

# in b.py
import logging
log = logging.getLogger(__name__)
class Bar(Foo):
  def do_something(self):
    log.info("clear in output it's logged from b")
```

If you want more granularity in special cases:
```python
# in c.py
import logging
log = logging.getLogger(__name__)
class Baz:
  def do_something(self):
    log.getChild('Baz').info("clear in output it's logged from c.Baz")
```

electrumx is pretty well organized though, I think most of the time the module `__name__` is enough to figure out where the log message is coming from; having `__class__.__name__` is a bit redundant if the class name and module name are about the same.

If you're okay with going to module level `log` I can do a PR for that along with the `Formatter` to simplify the module path.

eukreign:
@kyuupichan please let me know how you'd like me to proceed regarding above.

kyuupichan:
I still like the class name.  The reason is the code is currently organised functionally by class, and the purpose of it in the logs is not to locate the line giving the log but to understand wihch functional part of the code is emitting the message.

eukreign:
@kyuupichan `CompactFormatter` added in this PR: https://github.com/kyuupichan/electrumx/pull/518

PR #518, opened 2018-07-11
added compact log formatter that strips module name

eukreign:


kyuupichan:
Thanks for the approach.

Let me think about it; whatever I decide the heirarchical logs will be provided.

kyuupichan:
I've decided to go with this approach.  Unfortunately I renamed the file.  I want to move your logic into the library anyway (there's some other simple log logic there) so I'll create a separate commit.

Thanks for your suggestion.

kyuupichan:
Hmm, unfortunately the code doesn't work

kyuupichan:
Actually it does work.  There were 2 issues: I'd got confused renaming the file, and propagate needs to be set to False

PR #524, opened 2018-07-16
new protocol method: blockchain.transaction.get_id_from_pos

SomberNight:
For lightning, we need to be able to lookup short channel ids (block height, tx pos, output index) to transaction outputs.

I am not sure about the `tx_pos` name for the index inside the block, as the same string is used in `listunspent` with a different meaning; can rename if there is a better suggestion.

@ecdsa


docs/protocol-changes.rst
```diff
@@ -132,6 +132,11 @@ Changes
   * Optional *cp_height* argument added to
     :func:`blockchain.block.header` and :func:`blockchain.block.headers`
 
+New methods
+-----------
+
+  * :func:`blockchain.transaction.get_id_from_pos`
```
kyuupichan: Can you rename it to id_from_pos?  The initial "get" has been dropped from recent method names.

electrumx/server/controller.py
```diff
@@ -863,3 +863,21 @@ def mempool_get_fee_histogram(self):
...
+        hex_hashes = await self.daemon_request('block_hex_hashes', height, 1)
+        block_hash = hex_hashes[0]
+        block = await self.daemon_request('deserialised_block', block_hash)
+        tx_hashes = block['tx']
+        try:
+            tx_hash = tx_hashes[tx_pos]
+        except IndexError:
+            raise RPCError(BAD_REQUEST, f'no tx at position {tx_pos} in '
+                           f'block {block_hash} at height {height:,d}')
+        return tx_hash
```
kyuupichan: 60% of the logic here is common with transaction_get_merkle.  Can you break out a common async subroutine, perhaps tx_hashes_at_height?  It would validate height and do thhe daemon_request calls.  Finally also format tx_pos with ":,d" in the error message.

electrumx/server/session.py
```diff
@@ -533,6 +533,13 @@ def set_protocol_handlers(self, ptuple):
                 'blockchain.address.subscribe': self.address_subscribe,
             })
 
+        if ptuple >= (1, 4):
```
kyuupichan: Please merge the new handler into the existing 1.4 code.

kyuupichan:
FYI - I've been doing extensive restructuring of the session code.  Whilst there I created ElectrumX.block_tx_hash_list which should be useful for you (your code should go in session.py now).

PR #526, opened 2018-07-17
DeepOnion addition to coins.py and testblock

sean1000444:


sean1000444:
pip will not work for x13-hash, so I just noted in the coins.py file that a server ownder running a DeepOnion electrumx server will have to download it from source.

kyuupichan:
To add a coin to ElectrumX I want passing tests so somehow that package needs to be fixed.  It won't install for me on MacOSX with pip3

sean1000444:
That is just because for some reason the pip package won't work-otherwise it is perfectly fine. Anyone who runs an ElectrumX-DeepOnion server can just download the needed hash code from git as we note in the file. Would it be acceptable to comment out the parts that reference x13 in the codebase so the tests pass?

kyuupichan:
Have you raised the issue with the maintainers of the broken package?  That seems the best place to start

sean1000444:
I actually made it: originally there was only the git page for it.

kyuupichan:
Any progress here?  Or shall I close this?

PR #527, opened 2018-07-17
Fix sesion_count invocation

nijel:
This is broken since e41d1db2.


kyuupichan:
Thanks.  Your commit is conflicted (I've been changing a lot as you can see).  I fixed it separately (I hope) and credited you in the commit.

PR #531, opened 2018-07-23
fix mempool tx fee calculation typo

SomberNight:
Also, btw, in this commit https://github.com/kyuupichan/electrumx/commit/b05cc4e78bba4db39e7629f1ad20e82fd1ca86ff, the fee histogram recalculation period was changed from ~30 seconds to ~500 seconds.
Do you think it's that expensive? Would be nice to lower it back.


kyuupichan:
No I think that's fine, perhaps still frequent.  Why do you think it's insufficient?  Blocks are only every 10 mins...

PR #533, opened 2018-07-26
change BTX address format

cipig:
because of https://bitcore.cc/important-update-bitcore-public-keys/


kyuupichan:
Thanks.  In future, please add a changelog.rst entry too.

PR #540, opened 2018-07-29
Add Decred support

erasmospunk:


erasmospunk:
Is the test failure related to this pull request?

kyuupichan:
Travis enforces code style now:

$ pycodestyle electrumx/server/*.py electrumx/lib/*.py
electrumx/lib/coins.py:1726:28: E231 missing whitespace after ','

Thanks for the PR; it'll take a bit to digest

erasmospunk:
Let me fix the style fails and separate the 2 refactoring commits to another PR.

erasmospunk:
Updated. If you merge PR #541 this should have only one commit "Add Decred support".

kyuupichan:
OK thanks John.  I'll close this.  I don't think you need to do this b.t.w. - github provides a "squash" option on accepting a PR.

erasmospunk:
@kyuupichan the other PR contains only some general refactors, this contains the Decred integration.

kyuupichan:
Ah sorry I misunderstood; re-opened

electrumx/server/db.py
```diff
@@ -409,6 +410,10 @@ def lookup_hashXs():
             for each prevout.
             '''
             def lookup_hashX(tx_hash, tx_idx):
+                if tx_hash == TxInput.ZERO and tx_idx == TxInput.MINUS_1:
+                    # No hashX for coinbase-like inputs
+                    return None, None
+
```
kyuupichan: The changes look fine to me.  Can you explain this though?  This will be impacting utxo lookup for every other coin.  Is this just an optimization for Decred, or is it a necessary part of the semantics?  Is there some other way of achieving it?

kyuupichan: This code in db.py is not where I thought it was; I thought it was in a hot loop but it's not.

Can you add a comment about why this is needed?  It's not obvious on a casual read that it's Decred-specific.  Also I think I will simpify the is_coinbase code in tx.py to omit the tx hash comparison, the tx_idx comparison is enough.  Can you remove the tx_hash comparison from the above snippet?

I'll merge this with those 2 minor changes.  Thanks a lot.

electrumx/server/block_processor.py
```diff
@@ -256,13 +256,23 @@ def __init__(self, env, daemon, notifications):
             self.header_mc.truncate(self.height + 1)
         await self.prefetcher.reset_height(self.height)
 
-    async def reorg_hashes(self, count):
+    async def reorg_hashes(self, count=None):
```
kyuupichan: I don't think the default =None is needed - can we drop it and also in both instances of calc_reorg_hashes?

electrumx/server/db.py
erasmospunk: Removed completely this code.

electrumx/server/block_processor.py
erasmospunk: Done

erasmospunk:
The only issue that I have is the log spam:
```
INFO:MemPool:26 stragglers
INFO:MemPool:26 txs dropped
```
This seems to be caused by some missing UTXOs created by the vote/stake transactions that contain unknown op codes.

PR #544, opened 2018-07-31
Add Groestlcoin

Kefkius:
Add Groestlcoin mainnet and testnet.

If this way of accounting for Groestlcoin's transaction hash (single sha256 instead of double) is not acceptable, please advise.


electrumx/lib/tx.py
```diff
@@ -413,6 +413,48 @@ def read_header(self, height, static_header_size):
         return self._read_nbytes(header_len)
 
 
+class DeserializerGroestlcoin(DeserializerSegWit):
```
erasmospunk: Check the if using PR #548 will remove the need of copy-pasting code.

This should become something like this:
```
class DeserializerGroestlcoin(DeserializerSegWit):
    TX_HASH_FN = sha256
```

kyuupichan:
If #548 gets in please use that as per @erasmospunk.  Also new coins require block tests - see other examples

Kefkius:
I rebased onto the change and added block tests. Thanks for the shortcut, @erasmospunk!

PR #548, opened 2018-07-31
Allow custom transaction hashes

erasmospunk:


electrumx/lib/tx.py
```diff
@@ -101,7 +103,8 @@ def read_tx_and_hash(self):
         we process it in the natural serialized order.
         '''
         start = self.cursor
-        return self.read_tx(), double_sha256(self.binary[start:self.cursor])
+        return self.read_tx(), type(self).TX_HASH_FN(
```
kyuupichan: Why type(self) in these 3 lines?  just "self." should work fine.

I hope this doesn't have much impact on speed.

kyuupichan: John - you seem to be offline.  Thanks for the PR.
I'll make this change just using self.TX_HASH_FN which has your desired semantics and my tests show to run 75% faster in the interpreter

kyuupichan: Hmm, I'm seeing why you did this now, lol

kyuupichan: TX_HASH_FN = staticmethod(double_sha256) works; learn something every day.

I've pushed this.

kyuupichan:
See https://github.com/kyuupichan/electrumx/commit/4c7ab221a7f981b4f4c7ff8020a805c5bed21d12

PR #551, opened 2018-08-02
Fix groestlcoin address parsing

erasmospunk:
Is there a better way to conditionally import hash functions for coins without creating static methods?


kyuupichan:
Not really - we don't want everyone running ElectrumX to have to import 20 hash packages, so they will have to be implemented in coin-specific functions.

However, I don't see why this is on the TX object and not a COIN method?

erasmospunk:
Yes, we don't want to import hash functions unconditionally.

I meant to import them iff a specific coin is used.

> However, I don't see why this is on the TX object and not a COIN method?

As it needs `ENCODE_CHECK = partial(Base58.encode_check, hash_fn=...`, I couldn't find a way to use that hash function without importing it...

kyuupichan:
Thanks I've merged a tweaked version of this

PR #568, opened 2018-08-11
Pivx support

erasmospunk:


erasmospunk:
After rebasing to the latest version it broke due to the new mempool code. Will re-upload with a fix.

kyuupichan:
There's nothing here that's memo dependent. Travis is flaky and falling arbitrarily recently. I think this was one of those?

erasmospunk:
No, the problem was the zerocoin transactions that have coinbase-like inputs.

erasmospunk:
Anyway, I will try to fix it in a new PR.

kyuupichan:
OK - no tests failed, so if something is wrong I guess a new test is needed!

PR #574, opened 2018-08-14
Improve generation inputs handling

### LINKED ISSUE #570: Fix coins with coinbase/generation like transactions
There are coins like PIVX, Decred, Zcoin that use generation outpoints (hash=ZERO, idx=MINUS_1) in normal transactions.

In block processing this creates an issue with spending UTXOs but it is not noticed as those transactions are the same as generation transactions.

In recent versions of ElectrumX those generation-like transactions crash the mempool processor.

erasmospunk:
Fixes #570

### DIFF
--- a/electrumx/lib/tx.py
+++ b/electrumx/lib/tx.py
@@ -31,19 +31,22 @@ from collections import namedtuple
 
 from electrumx.lib.hash import sha256, double_sha256, hash_to_hex_str
 from electrumx.lib.util import (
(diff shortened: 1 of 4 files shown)


kyuupichan:
Can you add a test to the mempool tests that utlizes all the changes you had to make there?  i.e. it would fail if any were removed.  A comment on why max() is needed would be good.

kyuupichan:
Thanks very much!

PR #576, opened 2018-08-15
Added support for Bitg

cunhasb:
Added BitcoinGreen support for mainnet and testnet. Added testnet support for pivx and decred.


kyuupichan:
There's too much in one here. Please break it up into per-coin PRs

PR #578, opened 2018-08-16
Added support for Decred testnet

cunhasb:


kyuupichan:
Why does adding support for Decred testnet require changing 19 files and a huge changelog?

PR #580, opened 2018-08-16
Add support for Decred Testnet

cunhasb:


tests/test_blocks.py
```diff
@@ -73,4 +73,5 @@ def test_block(block_details):
     assert (coin.header_prevhash(block.header)
             == hex_str_to_hash(block_info['previousblockhash']))
     for n, (tx, txid) in enumerate(block.transactions):
-        assert txid == hex_str_to_hash(block_info['tx'][n])
+        if n < len(block_info['tx']):
```
kyuupichan: This change needs an explanation

setup.py
```diff
@@ -13,7 +13,7 @@
     # "xevan_hash" package is required to sync Xuez network.
     # "groestlcoin_hash" package is required to sync Groestlcoin network.
     install_requires=['aiorpcX>=0.7.1,<0.8', 'attrs',
-                      'plyvel', 'pylru', 'aiohttp >= 2'],
+                      'plyvel', 'pylru', 'aiohttp >= 2', 'blake256'],
```
kyuupichan: Sorry, no coin-specific packages can go here.  Please remove this and just add it to .travis.yml instead so it works for the tests.  But anyway - this seems to have nothing to do with decred testnet.

cunhasb: I just removed the 'blake256' from the setup.py, and also some styling errors on coin.py. Regarding your comment about nothing to do with decred testnet, the functionality still the same but the parameters are different for the testnet (XPUB_VERBYTES,XPRV_VERBYTES, etc...) is that what you are referencing to or was that blake256 on the setup.py?
Please let me know if you want me to do anything else.

tests/test_blocks.py
MatteBru: I think when Decred mainnet was added they combined normal and staking transactions (which are separate fields when retrieving block info from dcrd) into the block.json's tx array. Without doing that for the testnet block.json, this loop would go out of range on the ```block_info['tx']``` array because the decoded ```block.transactions``` include both staking and normal tx's.

kyuupichan:
This has conflicts now with the prior commit, otherwise it's good

PR #583, opened 2018-08-17
fix max session count limiting

SomberNight:
Max session count limiting was not working.

```
ERROR:asyncio:Exception in callback SessionManager._close_servers()
handle: <Handle SessionManager._close_servers()>
Traceback (most recent call last):
  File "/usr/lib/python3.6/asyncio/events.py", line 127, in _run
    self._callback(*self._args)
TypeError: 'coroutine' object is not callable
/usr/lib/python3.6/asyncio/base_events.py:284: RuntimeWarning: coroutine 'SessionManager._close_servers' was never awaited
  task = tasks.Task(coro, loop=self)
```

### DIFF
--- a/electrumx/server/session.py
+++ b/electrumx/server/session.py
@@ -575,7 +575,7 @@ class SessionManager(object):
                                 'connections until count drops to {:,d}'
                                 .format(self.max_sessions, self.low_watermark))
             loop = asyncio.get_event_loop()
-            loop.call_soon(self._close_servers(['TCP', 'SSL']))
+            asyncio.run_coroutine_threadsafe(self._close_servers(['TCP', 'SSL']), loop)
         gid = int(session.start_time - self.start_time) // 900
         if self.cur_group.gid != gid:
             self.cur_group = SessionGroup(gid)


electrumx/server/session.py
```diff
@@ -575,7 +575,7 @@ def add_session(self, session):
                                 'connections until count drops to {:,d}'
                                 .format(self.max_sessions, self.low_watermark))
             loop = asyncio.get_event_loop()
-            loop.call_soon(self._close_servers(['TCP', 'SSL']))
+            asyncio.run_coroutine_threadsafe(self._close_servers(['TCP', 'SSL']), loop)
```
kyuupichan: I'm not sure why this solves the problem you saw?

kyuupichan: I'm near-certain run_coroutine_threadsafe is not needed for anything in ElectrumX - superstition?

SomberNight: it solves the problem as `call_soon` expects a traditional method callback, while `run_coroutine_threadsafe` expects a coroutine

kyuupichan:
Ah I see.  No I don't want to use a method whose very name implies its use is to avoid threading issues.  There are no threading issues.

The real problem here is asyncio's unfortunate API choice to make things like connection_made() callback based and not asynchronous.  It limits what can reasonably be done in the function and doesn't admit pushback.

I need to think about how I want to solve this

PR #591, opened 2018-09-02
handle blocks which contain txes in non-causal order

markblundeberg:
This commit changes advance_txs() and backup_txs() so that instead of iterating over a block's transactions once, adding and removing utxos for each tx, the adding and removal steps are now done in separate loops. This allows the block processor to function just as well when the transactions are not ordered causally -- they may be put in any order.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,11 +1,9 @@
-.. image:: https://travis-ci.org/kyuupichan/electrumx.svg?branch=master
-    :target: https://travis-ci.org/kyuupichan/electrumx
-.. image:: https://coveralls.io/repos/github/kyuupichan/electrumx/badge.svg
-    :target: https://coveralls.io/github/kyuupichan/electrumx
 
 ===============================================
 ElectrumX - Reimplementation of electrum-server
 ===============================================
+
+*This version of ElectrumX has been slightly modified to support blocks with any order of transactions. It is compatible with both the Bitcoin ABC canonical transaction ordering that will activate in November 2018, and the traditional topological transaction ordering used in all other bitcoin consensus rules. Your ElectrumX server will identify as version **ElectronX**. Please submit bug reports relating to transaction ordering to [our repo](https://github.com/Electron-Cash/electrumx), otherwise send bug reports (and donations) to upstream author [Neil Booth](https://github.com/kyuupichan/electrumx).*
 
 For a future network with bigger blocks.
 


kyuupichan:
What is the impact of this on a sync to height 500k?  How have you checked the results are correct?

markblundeberg:
Hi Neil, so far I have only been testing this on a fork of BCH testnet that changes to non-causally ordered blocks. I have done several backups and advances (using e.g., `./electrumx_rpc reorg 1000`) past the fork point and it works well.

So far I haven't yet had the chance to test a full re-index of the testnet, to test performance differences; and I haven't at all tried it on mainnet.

(I anticipate what you'll say and I agree -- it certainly it needs more testing before it is merged.)

kyuupichan:
Well I'd like to see that is has no negative performance impact.  The reason I used lists rather than sets like you do is that's they're much more lightweight memory-wise.  That might be cancelled by the effect of not having to create the set from the list.

Also you now have 3 traversals of the transactions where there was only one before, which concerns me.

Is there any need for the reverse() in the backup function now?  That might not be necessary if the loop is broken into two.

Finally you need to fix the failures (overlong lines) and please don't comment out code; just remove it.

kyuupichan:
I think the spends is a mirage and probably because of the prefetcher not being optimally triggered

markblundeberg:
OK, my first test shows negligible difference in time -- I synced both new/old versions up to testnet block 601,417, using my laptop without running any other activity, and the results were:

    27m 17s -- new commit  6762f72d8cd7cb2b25c86fbd63256b990a3fef4d
    27m 25s -- old commit  2b123d77780d04644150ee0d8aa6838407285cf9

The new commit did manage to sync up to the tip with no issues (1h 20m), but strangely after block 601,417, the old commit ran out of memory (?!?) so I decided to just compare them up to that point. Around this block level, the testnet has some very annoying blocks that inflate the flush cache very rapidly. Still, I don't understand why the old commit would use so much more memory. Very strange.

I am running on a laptop with just 4 GB RAM and set CACHE_MB=500, and bitcoind is on a separate machine. I've noticed that ElectrumX's memory usage swings are way in excess of CACHE_MB (this is with Python 3.6.3), which may be worth looking into separately...

I re-checked the thing you mentioned with hashXs as sets/lists and I believe it shouldn't make a big difference, since they get deleted at the end of every block.

markblundeberg:
(that said, as you say with 21 MB blocks it could become a concern indeed)

markblundeberg:
FWIW in Python 3.6.3, sys.getsizeof on a list of 95000 elements gives me 805 KiB, and a set made from this list gives me 4096 kiB.

kyuupichan:
Yeah testnet is lumpy.  I'm more interested in mainnet.  The memory swings are mostly from prefetching N blocks, where N is determined from recent average block sizes, which are unstable on testnet (swinging from 1KB to 1MB) which doesn't happen on mainnet.  I had to limit N because of testnet; it used to go up to 4K blocks; now I think I capped it at 500.

markblundeberg:
OK I've tried the mainnet sync test you suggested. I set my bitcoind to stop syncing at height 100416 (January 1, 2011. I used a hard-fork parameter to do this, maybe there's a better way)

Electrumx sync time starting from empty db/ directory (NEW COMMIT): **60 seconds**

    INFO:Prefetcher:catching up to daemon height 100,416 (100,417 blocks behind)
    INFO:Prefetcher:verified genesis block with hash 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
    INFO:BlockProcessor:our height: 9 daemon: 100,416 UTXOs 0MB hist 0MB
    INFO:BlockProcessor:our height: 67,509 daemon: 100,416 UTXOs 11MB hist 21MB
    INFO:BlockProcessor:caught up to height 100416
    INFO:DB:flushed filesystem data in 0.03s
    INFO:History:flushed history in 0.7s for 176,359 addrs
    INFO:DB:flushed 100,417 blocks with 218,027 txs, 72,470 UTXO adds, 0 spends in 0.1s, committing...
    INFO:DB:flush #1 took 1.3s.  Height 100,416 txs: 218,027 (+218,027)
    INFO:DB:tx/sec since genesis: 3,608, since last flush: 3,609
    INFO:DB:sync time: 01m 00s  ETA: 11h 28m 04s
    INFO:BlockProcessor:ElectrumX 1.8.5 synced to height 100,416
    INFO:DB:closing DBs to re-open for serving

OLD COMMIT: **58 seconds**

    INFO:Prefetcher:catching up to daemon height 100,416 (100,417 blocks behind)
    INFO:Prefetcher:verified genesis block with hash 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
    INFO:BlockProcessor:our height: 9 daemon: 100,416 UTXOs 0MB hist 0MB
    INFO:BlockProcessor:our height: 70,009 daemon: 100,416 UTXOs 11MB hist 22MB
    INFO:BlockProcessor:caught up to height 100416
    INFO:DB:flushed filesystem data in 0.03s
    INFO:History:flushed history in 0.7s for 176,359 addrs
    INFO:DB:flushed 100,417 blocks with 218,027 txs, 72,470 UTXO adds, 0 spends in 0.1s, committing...
    INFO:DB:flush #1 took 1.3s.  Height 100,416 txs: 218,027 (+218,027)
    INFO:DB:tx/sec since genesis: 3,713, since last flush: 3,714
    INFO:DB:sync time: 58s  ETA: 11h 08m 37s
    INFO:BlockProcessor:ElectrumX 1.8.5 synced to height 100,416
    INFO:DB:closing DBs to re-open for serving

After each sync, I ran the following command querying the pizza address:

    ./electrumx_rpc query -l 100000 1XPTgDRhN8RFnzniWCddobD9iKZatrvH4 > query-full.txt

In both cases this creates a 343595 byte file with 3318 txes and 31 utxos; the contents are identical according to diff/cmp.

I'm going to try to try some syncs to a bigger height.

markblundeberg:
OK, I have now completed sync tests to block number 214568 (January 1 2013). I got 36m/38m for new/old commits, but then I noticed that my CPU was not being saturated, the culprit being that I have been syncing over a 3 MB/s link (wifi). This would have affected all previous test times as well, and I'm sure the wifi saturating would have caused significant variance in timing.

I then redid the tests with gigabit ethernet link, which kept `top`-reported CPU load around 90%-140%. So, the times for sync to block 214568, with freshly started bitcoind in each case:

    old commit: 17m 38s
    new commit: 17m 34s

Console outputs:
[run-old.log](https://github.com/kyuupichan/electrumx/files/2350601/run-old.log)
[run-new.log](https://github.com/kyuupichan/electrumx/files/2350600/run-new.log)

kyuupichan:
Thanks for confirming the data seems good.  I would like to try syncs to heights much closer to tip because until 200k blocks were small and not congested.  It gets interesting around 363k.

I want to think and analyse your approach; my main concern is being stuck with it long-term when blocks are bigger; it is more memory-hungry and might limit scalability

kyuupichan:
Can you re-submit so that the tests pass (over-long lines), you remove the commented line in history.py, and resolve the conflict(s)?  If I commit this I'd rather it have original authorship than mine.

markblundeberg:
Fair enough -- looking over it again, the change to sets seems optional, so it's always possible to go back to lists. For ultimate memory compression, you can have the hashXs as a compacted `bytearray` which would be 72 bytes plus 12 bytes per hashX entry. According to my calculations, the hashXs list and its bytearray items would be only 11 MB of objects, for that 21 MB block (95861 txes, 98462 + 192859 inputs+outputs). Actually less, since a good fraction of those outputs are op_returns.

I'll make those style changes to the code and see how far in height I can test the sync, with my meager hardware. Looks like it should be smooth enough, as you say the mainnet is less lumpy.

markblundeberg:
fixed style + rebased. I've also taken out the reversed() iterators in backup_txs and will try running this way for the next tests.

kyuupichan:
Yes it could be a byte array. However duplicates need to be removed so the awkwardness of handling at some point overwhelms the benefits of reduced memory usage.  Not sure where our if that happens though.  Thanks for your testing, I will get around to doing the same on hardware I have available.

electrumx/server/block_processor.py
```diff
@@ -402,34 +402,37 @@ def advance_txs(self, txs):
...
         undo_info_append = undo_info.append
         update_touched = self.touched.update
-        hashXs_by_tx = []
-        append_hashXs = hashXs_by_tx.append
 
-        for tx, tx_hash in txs:
-            hashXs = []
-            append_hashX = hashXs.append
-            tx_numb = s_pack('<I', tx_num)
+        hashXs_by_tx = [set() for _ in txs]
```
erasmospunk: Why not `hashXs_by_tx = [set() for _ in range(len(txs))]`?

erasmospunk:
A different approach: iterate the txs and try spending UTXOs like in the old method and in case of failure, add the input to a to-recheck list. Always add the outputs to the UTXO set. When all txs are processed, any out of order UTXO would be added to the DB so the to-recheck list would just spend the necessary UTXOs and raise an exception in case nothing is found.

markblundeberg:
I think that would in principle be fine for advancing. It might make backing up rather complicated though.

electrumx/server/block_processor.py
kyuupichan: I'm near certain you're right and that as-written is fastest

markblundeberg:
It looks like I may have to leave the rest of benchmarking to you @kyuupichan , I am starting to repeatedly get database corruption problems in my bitcoind -- likely my old hardware is starting to flake out.

WyseNynja:
I'm confused. Why was this closed? What does "seems to be for the new repo" mean?

kyuupichan:
You think a change to the name of the project was intended for this one?

markblundeberg:
@WyseNynja ah, it was my bad -- there is repo up at https://github.com/Electron-Cash/electrumx which has this any-order compatibility patch included. When I was setting up that repo I accidentally added commits to this pull request.

WyseNynja:
@kyuupichan I don't think the name change of the project was intended for this one. I do think the rest of the changes in this PR will be important depending on how the BCH fork goes in November though. What is your plan for supporting CTOR if it gets miner support?

markblundeberg:
@WyseNynja I just opened up a new PR at #633 since this one is permanently dead (not possible to change branch). IMO for now the status quo will work (the short term unofficial release supporting CTOR), so nothing is urgent.

kyuupichan:
@WyseNynja in the very unlikely event these changes get miner support I'll consider them.   However you should understand they come with a cost: higher memory requirements and hindrance of future scaling.  All coins should not be burdened with changes that only a fork of BCH needs.

PR #597, opened 2018-09-14
[WIP] [decred] Update to testnet3

dajohi:
While here, fix mainnet and testnet WIF_BYTE.


kyuupichan:
OK, I have no idea if the changes are good, but I see no reason not to trust you...

PR #610, opened 2018-10-05
Fix link to coins.py in documentation

FMCorz:
Fixing the link to `lib/coins.py` now under the `electrumx` folder.

### DIFF
--- a/docs/environment.rst
+++ b/docs/environment.rst
@@ -418,6 +418,5 @@ your available physical RAM:
 
   I do not recommend raising this above 2000.
 
-.. _lib/coins.py:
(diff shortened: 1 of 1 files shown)


kyuupichan:
Thanks but your patch doesn't work (breaks building the docs)

FMCorz:
Yes, I noticed. I'm not sure how we're meant to link to files outside of the `docs` folder, I'm not familiar with Sphinx. Although it seems that the problem was simpler than I initially thought, the patch was updated.

Somewhat related note: it seems that `sys.path.insert(0, os.path.abspath('..'))` in `conf.py` isn't needed.

PR #622, opened 2018-10-15
expose block polling delay

shyba:
Hi,
We are running lots of integration tests and 5 seconds is delaying it a lot, so I though of this change where we can set the polling time to a lower value during those tests. Could also be useful for coins with very low block time, but the main use I see is still testing.
I'm open to re-implementing it in any other way you find suitable.
Thank you


kyuupichan:
I don't want to change this or make it configurable; I will get too many reports of slow servers or non-responsiveness.  If you want to change it locally, you're free to do so.

eukreign:
@kyuupichan, would you please reconsider: there is no simple way to override this without duplicating the entire method.

kyuupichan:
Sorry I thought you'd put it in an envvar too

Issue #587, opened 2018-08-21
Namecoin and PUSHDATA indexing


JeremyRand:
Hi, I figured I'd reply to @kyuupichan 's [comment](https://github.com/namecoin/electrum-nmc/commit/c2b182c89656837dff0f549c74a72e4ab9f1701f#commitcomment-29931070) here so that we get better visibility (hope that's okay).

> Hi, I saw some of your writings here: https://namecoin.org/news/
> 
> I am thinking of changing ElectrumX to index not script hashes, but instead to index all (hashed) PUSH_DATA in scripts. This would still enable an Electrum client to look up its hash160 keys (in slightly greater generality), but would also mean I can index the various OP_RETURN things happening in BCH. Would an index based on PUSHHDATA solve the name script lookup issues you mention? I suspect indexed PUSH_DATA data would be enough for any client's requirements.

Yes, I'm pretty sure that should be fine.  Looking up name transactions by address would be done with identical PUSHDATA hashes as for currency transactions, which works fine.  And looking up name transactions by name would be done with the hash of the `name` PUSHDATA, which also works fine.  I suppose it would be possible for someone to deliberately spam the index by creating name transactions where the `value` PUSHDATA is equal to a name that someone else is watching... but I think this would be a quite expensive attack for not much volume of spam, and it's probably no worse in Namecoin than for Bitcoin.  Problems would probably happen if someone created a very short binary name (e.g. if someone picked the number 0x03 as their name, then they'd see lots of unwanted transactions that deal with 2-of-3 multisig), but I can't imagine that this would ever be a real-world use case.  So yeah, I think this would be a perfectly reasonable solution.

> Note: regarding your point about hashing scripts: it acutally provides no security / cover at all, unless the hash isn't found. Anyone can subscribe to a server for the tx history, get one tx, and find the address from that tx. For example.

Right, the main privacy use case here is specifically for the situation where someone wants to register a name but they don't want to reveal the name while they're looking up whether it's already registered (since if the ElectrumX server knows that a given name is about to be registered, the server operator could register it herself).  So it's precisely for the case where the hash isn't found that this matters.

Thanks for reaching out, and sorry for the delay getting back to you... your message arrived while I was away at a conference, and I've since been digging out of the work that accumulated while I was away.

kyuupichan:
@JeremyRand thanks for your feedback.  I'll close this issue

PR #626, opened 2018-10-19
Namecoin: Support indexing name scripts

JeremyRand:
Since the functionality discussed in https://github.com/kyuupichan/electrumx/issues/587 is not yet implemented, this PR enables indexing of name scripts (both by the address-determining part of the script and the name part of the script).  I've tested it with the master branch of Electrum-NMC.


JeremyRand:
Sorry about the Travis fail, I'll fix that ASAP

kyuupichan:
Hi Jeremy,

Can you add some tests that give at least 60% coverage of the new processing code in coins.py?  I may need to refactor this or other parts of ElectrumX in the future and it would be good to have tests so I am unlikely to break anything.

JeremyRand:
> Can you add some tests that give at least 60% coverage of the new processing code in coins.py? I may need to refactor this or other parts of ElectrumX in the future and it would be good to have tests so I am unlikely to break anything.

Yes, I can do that, hopefully later today.

JeremyRand:
Okay, so I think the only remaining code in this PR that isn't covered by tests is for functionality that also isn't tested for Bitcoin.  So, I think it's ready for review now.

PR #627, opened 2018-10-20
Add support for CivX mainnet and testnet

turcol:


electrumx/server/session.py
```diff
@@ -1437,3 +1437,24 @@ def get_payment_position(payment_queue, address):
...
+class RepBannerElectrumX(ElectrumX):
+    '''TCP server that handles incoming Electrum connections for some tokens.'''
+
+    async def replaced_banner(self, banner):
+        network_info = await self.daemon_request('getinfo')
+        ni_version = network_info['version']
+        daemon_version = (ni_version[:ni_version.find("-")]).replace("v","")
+        mmrb = daemon_version.split('.')
+        daemon_subversion = 0 if len(mmrb) < 4 else mmrb[3]
+
```
kyuupichan: I don't like this; if your coin's code is so old it doesn't have networkinfo, please change that in daemon.py to check for your coin and call a separate function that uses getinfo to return a dict with 'version' and 'subversion' 'members.  getinfo is obsolete/removed in other bitcoind's so I don't want to make a separate function for that which would give the impression it's supported.

kyuupichan:
You shouldn't be changing the banner function. It gets its info from networkinfo RPC. Change that instead

turcol:
There are coins that still don't have the networkinfo RPC implemented (I am working with three of them). 

In this case I will remove the class RepBannerElectrumX(ElectrumX) from session.py and the coins will only use the default banner message "You are connected to an electrumx server"

kyuupichan:
The build is failing with your patch

electrumx/server/daemon.py
```diff
@@ -445,3 +445,15 @@ def client_session(self):
         # FIXME allow self signed certificates
         connector = aiohttp.TCPConnector(verify_ssl=False)
         return aiohttp.ClientSession(connector=connector)
+
+
+class PreLegacyRPCDaemon(LegacyRPCDaemon):
```
kyuupichan: Any reason you inherit LegacyRPCDaemon rather than Daemon?

electrumx/server/daemon.py
```diff
@@ -445,3 +445,15 @@ def client_session(self):
...
+    '''Handles connections to a daemon at the given URL.
+
+    This class is useful for daemons that don't have the new 'getblock'
+    RPC call that returns the block in hex, and need the False parameter
+    for the getblock'''
+
+    async def deserialised_block(self, hex_hash):
+        '''Return the deserialised block with the given hex hash.'''
+        return await self._send_single('getblock', (hex_hash, False))
```
kyuupichan: So passing False returns the deserialised block which requires True for all other coins?  How did that happen?

electrumx/server/daemon.py
turcol: To manually recreate the block in bytes (as it is in the LegacyRPCDaemon)  as the getblock RPC call does not returns the block in hex

PR #633, opened 2018-11-05
BCH: use any-order block processor variant (coin-specific)

markblundeberg:
This is an updated version of #591, modified so that the any-order block processing is coin-specific.

It's perhaps getting a bit late to have any official release with this included, but I thought it would be worthwhile to make the PR to show the concept works well even in coin-specific form. Right now `testnet.imaginary.cash:50002` is running on this commit on top a testnet with CTOR activated early, via BU 1.5.0.0 (@imaginaryusername feel free to correct me).

Depending on how is the aftermath of Nov 15, it may make sense to modify the PR and make two different coins.py entries for the two chains, or, if only one chain survives then either adopt this PR as-is or abandon the PR.

@kyuupichan I can also easily alter this to use lists instead of sets for the hashXs, if memory consumption is still a concern.

(N.b., I have learned my lesson about doing a pull request from master like last time, so no weird contamination commits should show up on this one.)

### DIFF
--- a/electrumx/lib/coins.py
+++ b/electrumx/lib/coins.py
@@ -558,7 +558,6 @@ class BitcoinCashRegtest(BitcoinCashTestnet):
     PEERS = []
     TX_COUNT = 1
     TX_COUNT_HEIGHT = 1
-    BLOCK_PROCESSOR = block_proc.AnyOrderBlockProcessor
 
 
 class BitcoinSegwitTestnet(BitcoinTestnetMixin, Coin):
--- a/electrumx/lib/coins.py
+++ b/electrumx/lib/coins.py
@@ -370,6 +370,7 @@ class BitcoinCash(BitcoinMixin, Coin):
     TX_COUNT = 246362688
     TX_COUNT_HEIGHT = 511484
     TX_PER_BLOCK = 400
+    BLOCK_PROCESSOR = block_proc.AnyOrderBlockProcessor
     PEERS = [
         'electroncash.cascharia.com s50002',
         'bch.electrumx.cash s t',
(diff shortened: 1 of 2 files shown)


electrumx/lib/coins.py
```diff
@@ -556,6 +558,7 @@ class BitcoinCashRegtest(BitcoinCashTestnet):
     PEERS = []
     TX_COUNT = 1
     TX_COUNT_HEIGHT = 1
+    BLOCK_PROCESSOR = block_proc.AnyOrderBlockProcessor
```
SomberNight: looks redundant as it inherits from `BitcoinCashTestnet`

kyuupichan:
Sorry Mark, but BCH doesn't need this.  You're assuming the result.  A separate coin is OK (Deadalcoin?), but your changes increase memory requirements and hinder scaling.  If your chain takes over BCH I'll consider it after the hard fork.

discoltk:
@kyuupichan I'm trying to be as prepared as possible for the forks, despite the fact that I don't want either of them to happen.    I run BU mostly anyway, so that makes this a bit easier to manage.   If Electrum-x was critical path for me I'd certainly be needing the patch.   If the code is functional, what's the harm in allowing people to have the option?   Onegaishimasu...

kyuupichan:
I've said before - greater memory consumption and makes future scaling harder.

If you need their change see Electron-Cash.

markblundeberg:
The commit just added changes sets to lists, though I haven't had a chance to test it yet. I think that was the only concern with memory consumption?

kyuupichan:
I've merged your original code as BitcoinCashABC for the benefit of those who want to install / run a single codebase.  If your latest code is tested and more efficient please submit a PR.
FYI I will probably be making optimisations to normal block processing; I don't think they work for LTOR because of the non-linear processing.

markblundeberg:
Sweet! I believe the merged version was run live and tested by johoe so it should be good to go, unless something unforeseen slipped in. Makes sense to defer the untested alterations to another PR.

PR #644, opened 2018-11-13
Added NIX Support

phamels:
Add support for NIX (nixplatform.io)


kyuupichan:
Please add a test like other coins and fix the formatting issue(s) that break the build.

PR #659, opened 2018-11-29
Bitcoinabc regtest

ezegom:
Add BitcoinABC regtest to lib/coins.py


kyuupichan:
Please fix the failure.  Also why not inherit BitcoinCashABC and then no need for the block processor specification

ezegom:
I inherited from BitcoinTestnetMixin as it specifies the 4 address prefixes, WIF Byte, and RPC port. Seemed cleaner to inherit from this and only specify the block processor.

PR #666, opened 2018-12-08
Myriadcoin support

cryptapus:


kyuupichan:
Please add block tests like the other coins

PR #668, opened 2018-12-13
for Koto Sapling

wo01:
We will change the following things after Sapling upgrade.
* Header size
* MerkleRoot calculation


kyuupichan:
Sorry I don't want coin-specific code in session.py

wo01:
All right.
Thanks.

PR #670, opened 2018-12-16
Trezarcoin Support

ChekaZ:
Trezarcoin Integration comes with the Support of TxComment and Blake2s.


electrumx/lib/tx.py
```diff
@@ -379,6 +379,65 @@ def read_tx(self):
...
+        if version >= 2:
+            txcomment = self._read_varbytes()
+        else:
+            txcomment = b''
+        return TxTrezarcoin(version, time, inputs, outputs, locktime,
+                            txcomment)
+
+    @staticmethod
+    def blake2s_gen(data):
+        from hashlib import blake2s
```
kyuupichan: As these imports are part of the standard library they should go at the top of the file and not in the functions

electrumx/lib/tx.py
```diff
@@ -379,6 +379,65 @@ def read_tx(self):
...
+        _input112 = data + _full_merkle
+        _key = keyTwo + ntime + _nBits + _nonce + keyOne
+        '''Prepare 112Byte Header '''
+        blake2s_hash = blake2s(key=_key, digest_size=32)
+        blake2s_hash.update(_input112)
+        '''TrezarFlips - Only for Genesis'''
+        return "".join(map(str.__add__,
+                            blake2s_hash.hexdigest()[-2::-2],
+                            blake2s_hash.hexdigest()[-1::-2]))
+
```
kyuupichan: You need to fix the indentation here; it's failing the build

kyuupichan: Also what is the map(str.__add__) stuff?  These are binary bytes not strings?  I'm sure there's a simpler way to right that line

ChekaZ: HexDigest does give a string, where a flip of the hex needs to happen afterwards. Python isnt my ace game, so there surely might be better options to write that more simply.

electrumx/lib/tx.py
ChekaZ: Solved

aciddude:
@kyuupichan  @ChekaZ  - that's done now.

PR #673, opened 2018-12-18
chore: fix a markle merkle typo

m-schmoock:


SomberNight:
there are actually a few more of these here and there, just grep the codebase for "markle" :)

PR #681, opened 2019-01-03
Add compact_history.py script to installed scripts

### LINKED ISSUE #680: compact_history.py not included with pip install
I just ran into #185 and I'm running `compact_history.py` now. 

The problem is that I was pip installing from github and the compact_history.py script wasn't being included. I had to change my Dockerfile to download the code and then pip install it.

It would be nice if setup.py installed compact_history.py and any other additional scripts via entry_points or similar.

whizz:
Fixes #680


kyuupichan:
Can you remove the .py extension (and ensure it's marked executable) like the other scripts if we're going to do this?  Should it be renamed electrumx_compact_history?

kyuupichan:
Thanks

Issue #667, opened 2018-12-12
Truncate AuxPoW for headers that are covered by a checkpoint


JeremyRand:
AuxPoW data isn't relevant for headers covered by a checkpoint, since the AuxPoW data isn't needed to compute the block header hash, nor is it needed to verify that the checkpoint commits to the block header.  (I'm assuming that our threat model excludes the possibility of the checkpoint committing to invalid blocks.)  Given this, we could save a lot of bandwidth by truncating AuxPoW data for headers that are covered by a checkpoint (i.e. for headers whose height is less than or equal to `cp_height` in `blockchain.block.header` and `blockchain.block.headers`).  In Namecoin's case, that means truncating everything after the first 80 bytes of each header; I'm not certain whether other AuxPoW coins have a different definition (though I suspect it's likely that some do).

Would a PR be accepted that truncates the AuxPoW data for headers that are covered by a checkpoint?  (The Namecoin devs would probably be willing to code up such a PR if no one else does it first.)

JeremyRand:
CC'ing @erasmospunk since he NACKed a superficially similar change in https://github.com/kyuupichan/electrumx/pull/457#discussion_r185293088 and I don't want to appear to step on any toes.

kyuupichan:
This sounds like an API change?  Or am I misunderstanding?  The server has no checkpoints; they are all client-side.

JeremyRand:
> This sounds like an API change? Or am I misunderstanding? The server has no checkpoints; they are all client-side.

@kyuupichan my issue said "for headers whose height is less than or equal to `cp_height` in `blockchain.block.header` and `blockchain.block.headers`"; that's how I'm defining headers that are covered by a checkpoint.  Sorry if that was unclear.

kyuupichan:
Oh I see, so it would change the nature of the returned value.  Got it.  That would probably need a protocol version bump

JeremyRand:
> That would probably need a protocol version bump

Agreed.  I'm actually not familiar with what's involved in doing a protocol version bump, as I've never submitted one to ElectrumX before... anything notable about what's involved there?

kyuupichan:
Just to update the docs to describe the changes and the three places in the code that refer to the version (setup.py, __init__.py and docs/make.conf).  Since this is a minor change I think it should be 1.4.1 rather than 1.5

JeremyRand:
As discussed over on #namecoin-dev, @ahmedbodi is possibly interested in implementing this.

[...]

PR #685, opened 2019-01-10
Viacoin: truncate auxpow from headers

romanornr:
Issue: https://github.com/kyuupichan/electrumx/issues/667

One of our devs fixed this issue a few months ago. We had the same issue.
We had our own fork where we truncated the headers
(Done in march 2018: https://github.com/reorder/electrumx/commit/0bfadbf27c4854aa083b97f6ef7ee34e1c770176)

So https://github.com/kyuupichan/electrumx/issues/667#issuecomment-447657576
This should be the fix.

### DIFF
--- a/electrumx/lib/coins.py
+++ b/electrumx/lib/coins.py
@@ -680,6 +680,19 @@ class Viacoin(AuxPowMixin, Coin):
         'viax1.bitops.me s t',
     ]
 
+    # Auxpow is redundant in SPV, truncate it from headers
+    @classmethod
+    def block_header(cls, block, height):
+        return super().block_header(block, height)[:cls.BASIC_HEADER_SIZE]
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


kyuupichan:
Thanks.  Is a simple test for this possible?

JeremyRand:
I don't think this behavior is correct.  AuxPoW is definitely not redundant in SPV, and it shouldn't be removed for headers where SPV verification is desired.  It is only redundant in headers that are covered by a checkpoint, since no PoW verification is needed for such headers.

JeremyRand:
Also note that @erasmospunk already NACKed this in https://github.com/kyuupichan/electrumx/pull/457#pullrequestreview-116651853 for the same reason that I just stated.

kyuupichan:
OK closing this, thanks

Issue #684, opened 2019-01-10
ERROR:root:task crashed NameError: name 'electrum' is not defined


Folkaio:
Hello,

I was getting the error ERROR:root:task crashed after updating from version 1.8.12 to 1.9.0.

It seems to be fixed after replacing this line : 
`self.hsub_results = (electrum, {'hex': raw.hex(), 'height': height})`
by
`self.hsub_results = (electrumx, {'hex': raw.hex(), 'height': height})`

in the file **/usr/local/lib/python3.6/dist-packages/electrumx/server/session.py**

ERROR output
----------------
electrumx_server[31820]: ERROR:root:task crashed: <Task finished coro=<Controller.serve() done, defined at /usr/local/lib/python3.6/dist-packages/electrumx/server/controller.py:81> exception=NameError("name 'electrum' is not defined",)>
electrumx_server[31820]: Traceback (most recent call last):
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/aiorpcx/util.py", line 118, in check_task
electrumx_server[31820]:     task.result()
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/electrumx/server/controller.py", line 134, in serve
electrumx_server[31820]:     await group.spawn(wait_for_catchup())
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/aiorpcx/curio.py", line 231, in __aexit__
electrumx_server[31820]:     await self.join()
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/aiorpcx/curio.py", line 200, in join
electrumx_server[31820]:     raise task.exception()
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/electrumx/server/session.py", line 498, in serve
electrumx_server[31820]:     await notifications.start(self.db.db_height, self._notify_sessions)
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/electrumx/server/controller.py", line 63, in start
electrumx_server[31820]:     await self.notify(height, set())
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/electrumx/server/session.py", line 554, in _notify_sessions
electrumx_server[31820]:     await self._refresh_hsub_results(height)
electrumx_server[31820]:   File "/usr/local/lib/python3.6/dist-packages/electrumx/server/session.py", line 345, in _refresh_hsub_results
[...]

PR #686, opened 2019-01-10
Fix NameError: name 'electrum' is not defined

romanornr:
Should fix https://github.com/kyuupichan/electrumx/issues/684


kyuupichan:
Lol, electrumx is a python module; this isn't the right fix.  Sorry for the bug; I'll fix it with a new release in a few minutes.

PR #694, opened 2019-01-15
add bitcoin cash testnet

### LINKED ISSUE #652: BCH testnet servers
Looks like testnet servers will have to be figured out too. I'm not quite sure what the correct form of inclusion is here, so I'll put this as an issue rather than PR.

Currently I know of the following on BU/ABC post-fork chain:

* bch0.kister.net ssl 51002
* testnet.imaginary.cash ssl 50002
* blackie.c3-soft.com ssl 60002

This one is on ABC 0.17 chain:

* electrum-testnet-abc.criptolayer.net ssl 50112

Not sure if there are any on SV chain, but I suppose peer discovery would not show them?

markblundeberg:
fix #652 

I put `NAME = "BitcoinCashABC"` for consistency with the other ones.

### DIFF
--- a/electrumx/lib/coins.py
+++ b/electrumx/lib/coins.py
@@ -539,6 +539,17 @@ class BitcoinSVTestnet(BitcoinTestnetMixin, Coin):
     ]
 
 
+class BitcoinCashTestnet(BitcoinTestnetMixin, Coin):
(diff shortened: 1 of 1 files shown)


kyuupichan:
Thanks

PR #700, opened 2019-01-17
Peercoin tests

peerchemist:


kyuupichan:
This causes the tests to fail

PR #711, opened 2019-01-26
Increase MAX_SEND for AuxPoW chains

JeremyRand:
AuxPoW headers are significantly larger than the headers in Bitcoin, so the current default MAX_SEND of 1 MB is insufficient.  In Namecoin mainnet, 5 MB wasn't enough to sync, while 10 MB worked fine.  I suspect that most other AuxPoW chains will be similar, so 10 MB seems like a decent default for AuxPoW chains.

### DIFF
--- a/docs/environment.rst
+++ b/docs/environment.rst
@@ -228,9 +228,10 @@ raise them.
 .. envvar:: MAX_SEND
 
   The maximum size of a response message to send over the wire, in
-  bytes.  Defaults to 1,000,000.  Values smaller than 350,000 are
-  taken as 350,000 because standard Electrum protocol header "chunk"
-  requests are almost that large.
+  bytes.  Defaults to 1,000,000 (except for AuxPoW coins, which default
+  to 10,000,000).  Values smaller than 350,000 are taken as 350,000
+  because standard Electrum protocol header "chunk" requests are almost
+  that large.
 
   The Electrum protocol has a flaw in that address histories must be
   served all at once or not at all, an obvious avenue for abuse.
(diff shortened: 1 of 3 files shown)


JeremyRand:
I have no idea what the Travis fail is caused by; it appears to be in a completely unrelated part of the code to what this PR touches.  Given that it only shows up in Python nightly, my hunch is that it's a Python bug, but I'm not entirely sure.

kyuupichan:
Can you also update the documentation please?

I'd prefer to make DEFAULT_MAX_SEND a coin property.  What do you think of that?  If the base class has 1 MB as now, AuxPowMixin could redefine it to 10MB.  Or each derived class could; doing it in the mixin might make it sensitive to multiple-inheritance order.

JeremyRand:
> Can you also update the documentation please?

Good point, I'll do that.

> I'd prefer to make DEFAULT_MAX_SEND a coin property. What do you think of that? If the base class has 1 MB as now, AuxPowMixin could redefine it to 10MB.

Yes, that does indeed sound cleaner than the check that this PR currently does.

> Or each derived class could; doing it in the mixin might make it sensitive to multiple-inheritance order.

True.  I guess I'll set each of the AuxPoW coin classes to 10 MB, since I assume that's a reasonably sane default for them, and if anyone who uses those coins finds that it's not optimal, they can submit another PR that refines it.

I'll see if I can update this PR tomorrow, though it might need to wait till Monday.

JeremyRand:
PR updated.

> Or each derived class could; doing it in the mixin might make it sensitive to multiple-inheritance order.

@kyuupichan Can you elaborate on exactly what failure mode would happen here?  AFAICT the `AuxPoWMixin` class already overrides several attributes of `Coin` without any bad effects.

kyuupichan:
There may be nothing to worry about indeed

Issue #713, opened 2019-02-02
MemoryError('dropping message over 1,000,000 bytes and re-synchronizing',)


sumBTC:
I'm running Electrumx since a long time now and continuously monitor the number of peers. Yesterday I suddenly got strange values. Turned out:

`electrumx_rpc peers`

gave me 

`MemoryError('dropping message over 1,000,000 bytes and re-synchronizing',)`

A restart didn't help. I found out the solution was to add 

`MAX_SEND=10000000`

(default MAX_SEND=1000000)

to 

/etc/electrumx.conf

Electrumx is now running again without problems.

I decided to submit this issue as it might be useful to others.

Question is: shouldn't Electrumx exit with an error message telling what went wrong and how to solve it?

Anyway, a big thank you to all the developers for all the hard work you are doing. Awesome project!

System information:

Ubuntu: 16.04
Electrumx:  1.9.1
COIN: BitcoinSegwit

More info on my [website](http://Bitkoins.nl).

kyuupichan:
I believe the typo just fixed above should fix this issue; let me know...

sumBTC:
I removed the `MAX_SEND=10000000` and the problem is gone. The only thing is: I'm not able to reproduce the issue but if there is a typo that has to be fixed anyway and if you're sure it has something to do with the issue, it's probably solved by this.

Again, thanks for all your efforts.

shsmith:
I think this is related to the ongoing Sybil attack against ElectrumX servers. Recently there have been just under 100 legit servers, but in the last 24 hours we had over 248 servers, which probably exceeded the hardcoded limit. Enough of the malicious servers have gone stale that we are back below the limit. Malicious servers are in the majority now, especially for older vulnerable clients that don't understand protocol version 1.4.

kyuupichan:
My change fixed this; there is not intended to be a limit on local RPC size

SomberNight:
Please reopen. The fix is insufficient. The issue still occurs.

SomberNight:
To test, patch:
```patch
diff --git a/electrumx/lib/text.py b/electrumx/lib/text.py
index 8546eb9..24ce76e 100644
--- a/electrumx/lib/text.py
+++ b/electrumx/lib/text.py
@@ -52,6 +52,7 @@ def peers_lines(data):
     '''A generator returning lines for a list of peers.
 
     data is the return value of rpc_peers().'''
+    return data
     def time_fmt(t):
         if not t:
             return 'Never'
[...]

PR #722, opened 2019-02-06
LocalRPC: don't limit message size

SomberNight:
fixes https://github.com/kyuupichan/electrumx/issues/713

needs https://github.com/kyuupichan/aiorpcX/pull/13

### DIFF
--- a/electrumx_rpc
+++ b/electrumx_rpc
@@ -115,6 +115,7 @@ def main():
         try:
             async with timeout_after(15):
                 async with Connector(RPCSession, 'localhost', port) as session:
+                    session.framer.max_size = 0
                     result = await session.send_request(method, args)
                     if method in ('query', ):
                         for line in result:


electrumx/server/session.py
```diff
@@ -1291,6 +1291,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.client = 'RPC'
         self.connection.max_response_size = 0
+        self.framer.max_size = 0
```
kyuupichan: I'm not sure about this - I think we only want to have no limit in the electrumx_rpc client no?  I don't see a reason for unlimited incoming message size.

SomberNight: Oh yes, you're right. For some reason I was thinking the limit is problematic both ways; but actually the framer is only used for incoming messages, so this line is not needed. I'll remove.

kyuupichan:
I've released 1.9.4 with the 2nd chunk only of this.

PR #723, opened 2019-02-07
Blacklist logic:

ecdsa:
* do not return blacklisted peers to clients
* blacklist is retrieved from BLACKLIST_URL
* blacklist is refreshed every 10 minutes
* for BitcoinSegwit, BLACKLIST_URL defaults to https://electrum.org/blacklist.json

### DIFF
--- a/electrumx/server/env.py
+++ b/electrumx/server/env.py
@@ -75,6 +75,10 @@ class Env(EnvBase):
         self.bandwidth_limit = self.integer('BANDWIDTH_LIMIT', 2000000)
         self.session_timeout = self.integer('SESSION_TIMEOUT', 600)
         self.drop_client = self.custom("DROP_CLIENT", None, re.compile)
+        self.blacklist_url = self.default('BLACKLIST_URL', None)
+        # temporary default
+        if self.blacklist_url is None and self.coin.NAME == 'BitcoinSegwit':
+            self.blacklist_url = 'https://electrum.org/blacklist.json'
 
         # Identities
         clearnet_identity = self.clearnet_identity()
--- a/electrumx/server/peers.py
+++ b/electrumx/server/peers.py
@@ -12,6 +12,8 @@ import random
 import socket
 import ssl
 import time
+import json
+import aiohttp
 from collections import defaultdict, Counter
 
 from aiorpcx import (Connector, RPCSession, SOCKSProxy,
@@ -75,6 +77,8 @@ class PeerManager(object):
         self.permit_onion_peer_time = time.time()
         self.proxy = None
         self.group = TaskGroup()
+        # refreshed
+        self.blacklist = set()
 
     def _my_clearnet_peer(self):
         '''Returns the clearnet peer representing this server, if any.'''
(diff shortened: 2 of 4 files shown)


cculianu:
I like this approach. 

Side note/question: What's the current "potential sybil peer" logic doing? I keep getting lots of false positives on it and I'm nervous any serious attacker would be able to trick any heuristic method.

This method, where a trusted central server provides a list of bad peers.. is definitely the way to go, IMHO.

jkister:
is it worth having the refresh time pulled from blacklist.json itself rather than being hardcoded?

electrumx/server/env.py
```diff
@@ -75,6 +75,10 @@ def __init__(self, coin=None):
         self.bandwidth_limit = self.integer('BANDWIDTH_LIMIT', 2000000)
         self.session_timeout = self.integer('SESSION_TIMEOUT', 600)
         self.drop_client = self.custom("DROP_CLIENT", None, re.compile)
+        self.blacklist_url = self.default('BLACKLIST_URL', None)
+        # temporary default
+        if self.blacklist_url is None and self.coin.NAME == 'BitcoinSegwit':
+            self.blacklist_url = 'https://electrum.org/blacklist.json'
```
cculianu: Would a `coins.py` member such as `BitcoinSegwit.FALLBACK_BLACKLIST_URL_FALLBACK` or whatever also handle this gracefully?

electrumx/server/peers.py
```diff
@@ -477,6 +506,12 @@ def on_peers_subscribe(self, is_tor):
                   not peer.bad and peer.is_public]
         onion_peers = []
 
+        if not is_peer:
+            recent = filter(self._is_allowed, recent)
+            recent = [peer for peer in self.peers
+                      if self._is_allowed(peer)
+                      and peer.host in self.sybils]
```
kyuupichan: This doesn't look right

electrumx/server/peers.py
```diff
@@ -129,6 +134,29 @@ def _permit_new_onion_peer(self):
...
+            return
+        while True:
+            try:
+                async with session.get(url) as response:
+                    r = await response.text()
+                    self.blacklist = set(json.loads(r))
+                    self.logger.info('blacklist retrieved from "%s": %d'
+                                     % (url, len(self.blacklist)))
+            except Exception as e:
+                self.logger.info('could not retrieve blacklist, "%s"' % url)
```
kyuupichan: Should probably be an error

electrumx/server/peers.py
```diff
@@ -75,6 +77,9 @@ def __init__(self, env, db):
         self.permit_onion_peer_time = time.time()
         self.proxy = None
         self.group = TaskGroup()
+        # refreshed
+        self.blacklist = set()
+        self.sybils = set()
```
kyuupichan: I'll get rid of the sybils in a follow-up I think

kyuupichan:
I've pushed a follow-up; let me know if it works please

ecdsa:
it looks good to me

ecdsa:
thanks for merging

PR #729, opened 2019-02-09
add .idea dir to gitignore

RaySpock:


kyuupichan:
There is no such directory in the repo

RaySpock:
I'm using PyCharm as my IDE and all the project specific setting will be stored under this folder(Just like .vscode). That's why I want to make this change to let workspace cleaner. But thanks for taking the time to review it.

kyuupichan:
Fair enough; thanks for explaining.  I've merged this.

PR #738, opened 2019-02-22
Add support for Bitsend (BSD)

dalijolijo:


setup.py
```diff
@@ -15,7 +15,7 @@
     # "groestlcoin_hash" package is required to sync Groestlcoin network.
     # "pycryptodomex" package is required to sync SmartCash network.
     install_requires=['aiorpcX>=0.10.4,<0.11', 'attrs',
-                      'plyvel', 'pylru', 'aiohttp >= 2'],
+                      'plyvel', 'pylru', 'aiohttp >= 2', 'xevan_hash', 'x11_hash'],
```
kyuupichan: No coin-specific modules thanks

dalijolijo: How should I define the coin-specific modules? Only as comment in setzp.py?

kyuupichan: If you want.  Whoever runs a server for your coin needs to install them, but we shouldn't be reuqiring anyone who runs any server to install all the python hash modules of every other coin out there

dalijolijo:
> Also, the tests are failing

fixed

Issue #740, opened 2019-02-26
Implementing another connection mechanism


eabz:
Hello @kyuupichan 
We are trying to use electrumx server as the main UTXO tracker for our app. 
The thing is, there is no way to connect an Ionic app to electrumx using TCP. And we want the user to be able to connect to any electrum node.
Is there any easy way to create another connection mechanism? We can use any other, http, websocket, rpc...

kyuupichan:
https://github.com/kyuupichan/electrumx/issues/499

Lots of people ask, no one does the work...

eabz:
@kyuupichan 
True... 
I am not good at python, but tell me. Should we work directly on electrumx or on aiorpcX.
Also, is websockets a good idea?

kyuupichan:
Yes I think websockets would be useful; there is an excellent async Python package for that.  I'm unclear if it's better in aiorpcX or ElectrumX; don't worry too much about that

kyuupichan:
I just implemented websockets in https://github.com/kyuupichan/electrumx/commit/a6bfc032674aac71eeffea3457ad10c456c1baa0

See https://electrumx.readthedocs.io/en/latest/environment.html#services

Issue #499, opened 2018-06-19
Websocket support?


DaniGuardiola:
Hi, I would like to know if there's websocket support, it would make things much easier for my lightweight wallet. If there is, how do I use it and what are the differences with a standard socket connection? Is there any documentation available? If there isn't, could it be implemented? Is it planned? Would you be open to a PR if you need it? I could get someone to work on it.

Thank you in advance :)

kyuupichan:
See #242 perhaps?  I am not working on these, but will accept a quality PR with tests to implement such a feature.

DaniGuardiola:
@kyuupichan thanks for your response!

Good to know that you will accept a quality PR. Could you write down some basic guidance on where to start and what to check in the code in order to implement this?

Thank you :+1:

kyuupichan:
I don't really know much about websockets.  The code should be asynchronous.  aiohttp might be useful.  It's not clear to me if it should use or be part of the aiorpcX package.

DaniGuardiola:
Where are the TCP sockets currently implemented? It could be a good reference to understand how to provide the websocket connection as well.

DaniGuardiola:
I made this demo / info repo, could you take a look at it and let me know if you have any suggestion? @kyuupichan 

Thanks a lot, I just want to double-check that I have correct and complete information before starting development.

kyuupichan:
Looks like the right idea to me.  Note I'm unsure if the appropriate location for the new code is aiorpcX or electrumX.  Probably the former.  Note aiorpcX has 100% test coverage.

kyuupichan:
I think the websockets package is the right way to do this.  It shouldn't be more than a few lines of code added in aiorpcx.  I'll take a look at it some time in the next month or so.

DaniGuardiola:
@kyuupichan thanks, I've abandoned the project that depended on this, but on the other hand, I've been working with Python so I should be more prepared to help. I'll see if I can find some time to help you out with the PR.

kyuupichan:
I just implemented websockets in https://github.com/kyuupichan/electrumx/commit/a6bfc032674aac71eeffea3457ad10c456c1baa0

See https://electrumx.readthedocs.io/en/latest/environment.html#services

DaniGuardiola:
Thanks a lot @kyuupichan!! Nice work :D

kyuupichan:
@DaniGuardiola have you tried it and confirm it works as you expect?

[...]

PR #741, opened 2019-02-26
Initial WS Server

eabz:
Reference: #740, #499 #242 
@kyuupichan 
What do you think? Is very elemental, but is working. 
At least is working with my app


erasmospunk:
Is it possible to add some tests?

eabz:
@erasmospunk 
What kind of tests?

.travis.yml
```diff
@@ -26,7 +26,9 @@ install:
   - pip install pytest-asyncio
   - pip install pytest-cov
   - pip install Sphinx
-# hashes
+  - pip install websockets
+  - pip install jsonrpcserver
```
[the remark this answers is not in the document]
eabz: @kyuupichan Removed jsonrpcserver

eabz:
@kyuupichan 
Remove jsonrpcserver and revert to websockets library.
Also, using ElectrumX all methods are available on websockets

electrumx/server/session.py
```diff
@@ -142,6 +143,39 @@ def __init__(self, env, db, bp, daemon, mempool, shutdown_event):
         LocalRPC.request_handlers = {cmd: getattr(self, 'rpc_' + cmd)
                                      for cmd in cmds}
 
+    async def _start_ws_server(self, *args, **kw_args):
+        instance = ElectrumX(self, self.db, self.mempool, self.peer_mgr, 'TCP')
+
+        @asyncio.coroutine
```
kyuupichan: The decorator is not necessary; it appears nowhere else in ElectrumX.

electrumx/server/env.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, coin=None):
         self.reorg_limit = self.integer('REORG_LIMIT', self.coin.REORG_LIMIT)
         # Server stuff
         self.tcp_port = self.integer('TCP_PORT', None)
+        self.websocket_port = self.integer('WEBSOCKET_PORT', None)
```
kyuupichan: It's not clear if this is SSL or plain text communication.

Can we support both?  Perhaps with env vars WS_PORT and WSS_PORT?

kyuupichan:
This is looking promising, thank you; give me a bit of time to review.  Please also see the 2 comments I just left.

I don't have client code to test with - can you confirm it works for you?

kyuupichan:
Also please update the documentation for the envvar(s) you add.

electrumx/server/session.py
eabz: Removed

electrumx/server/env.py
eabz: @kyuupichan I think the best option is check for SSL_CERTFILE and SSL_KEY, if those are specified, ElectrumX will serve a Secure WebSocket

eabz: Updated

eabz:
@kyuupichan 
Added Secure WebSocket support and docs.
The Server is working, I am testing it against an app created on Ionic. 
I am still testing the WSS.

eabz:
@kyuupichan 
Hello, I have tested it all weekend and is working. 
Currently supports:
- Websockets ("ws://")
- Secure websockets ("wss://")
- Version handshake
- Single querys
- Batches.

kyuupichan:
Thanks for confirming.  Give me some time to review please.

eabz:
@kyuupichan 
Hello, do you get a chance to review the PR?

kyuupichan:
Thanks for this.  I've just taken a look.  The fact you have to worry anything about JSON means that it's not being abstracted correctly.  I think I need to rework the session class in aiorpcX, then integrating websockets should be easy.

eabz:
Thanks @kyuupichan we are currently reworking our backend and we may don't use electrumx anymore, if you can take it from here it will be ok, but I am not working on it further. If this is a good start for you is perfect, but if don't, you can close it.

kyuupichan:
I'll be making the session class independent of asyncio protocols; then it can simply having a transport.  That can be an asyncio one (suitably inherited) or it can be a websocket one (using the package).  Then in ElectrumX it will just be a simple change to add websocket support once I update it for the new interface.

I may copy some of your documentation for the new env vars, thanks.

kyuupichan:
I just implemented websockets in https://github.com/kyuupichan/electrumx/commit/a6bfc032674aac71eeffea3457ad10c456c1baa0

See https://electrumx.readthedocs.io/en/latest/environment.html#services

Note how none of the server code really needed to change

Issue #746, opened 2019-03-03
Should blacklisted peers still be visible in "electrumx_rpc peers" command?


sumBTC:
I'm running Electrumx 1.9.5 (COIN = BitcoinSegwit). 

If I understand it well, peers are automatically blacklisted if they appear in https://electrum.org/blacklist.json. One of those is "*.electrupdate.pro" but when I do `electrumx_rpc peers` they are still visible. Is it normal that blacklisted peers are still visible with this command?

sumBTC:
Oh, they have label "bad". Closing issue and sorry.

peleion:
Yes:
https://github.com/spesmilo/electrum/issues/5091#issuecomment-464345295

sumBTC:
Thanks for the answer. I'll close this issue.

sumBTC:
Issue is solved in #753.

PR #753, opened 2019-03-09
peers: mark blacklisted peers bad

SomberNight:
My server has ~9.5k total peers, of which 1.5k is "good", on BTC mainnet.
I also observe some general slowness and fewer than usual session count; which I am not sure if related to the absurd peer count, but it might be.

I think blacklisted peers should be marked bad, so we don't gossip about them to other peers, and we don't keep monitoring them; which is what this PR does.

@ecdsa what do you think?

### DIFF
--- a/electrumx/server/peers.py
+++ b/electrumx/server/peers.py
@@ -148,6 +148,10 @@ class PeerManager(object):
                 self.logger.error(f'could not retrieve blacklist from {url}: {e}')
             await sleep(600)
 
+    def _is_blacklisted(self, host):
+        return any(item in self.blacklist
+                   for item in (host, '*.' + '.'.join(host.split('.')[-2:])))
+
     async def _detect_proxy(self):
         '''Detect a proxy if we don't have one and some time has passed since
         the last attempt.
@@ -296,6 +300,9 @@ class PeerManager(object):
         return False
 
     async def _verify_peer(self, session, peer):
+        if self._is_blacklisted(peer.host):
+            raise BadPeerError('blacklisted')
+
         if not peer.is_tor:
             address = session.peer_address()
             if address:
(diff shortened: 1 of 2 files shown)


ecdsa:
this looks good to me (I have not tested it though)

sumBTC:
Does this solve issue #746?

sumBTC:
I'm running with this change. Solves #746.

freerko:
I updated my servers and at first it seemed like this is working as expected, I got around 90 good peers. But after a few hours it jumped to around 140 with about 50 new *.electrumservice.com.co peers marked as good. According to a passive DNS database those hostnames first appeared only a few hours ago. But since *.com.co is on the blacklist.json already they shouldn't appear as good peers. After a restart of electrumx they don't appear in the peers list anymore. 
Does this commit (or the whole blacklist logic) correctly handle new peers that are first showing up only during runtime and haven't been known before?

SomberNight:
@freerko Every 10 mins, the blacklist is refreshed.
Every time a client/other server asks for our peerlist, we only send peers that are not blacklisted.

However peers are only marked bad when we (re-)test them, in `_monitor_peer`. This test will happen when we first get to know about them, and then, if they are good, we keep monitoring them, and re-test every ~24 hours.

So, what you describe can indeed happen, it happens for "new" evil peers:
- server learns about evil peer attacker newly set up
- server runs basic tests (same chain, not lagging much, not on blacklist, ...)
- as it's a newly set up server, it will not be on the blacklist yet
- server propagates around peer discovery
- we learn about it, and blacklist it
- every 10 mins, servers redownload the blacklist
- the next time `_monitor_peer` checks the peer, it will mark it bad (but this check is every ~24 hours)

freerko:
Thanks for the explanation. 
If *.com.co has already been on the blacklist.json for more than about 10 hours the 50 new *.electrumservice.com.co should have never made it past the second step you mentioned into my "good" peer list. If *.com.co was added only today to the blacklist I understand that it is possible that I got the new evil peers before that.

Issue #756, opened 2019-03-13
Eliminate peers listing multiple DNS entries for same IP  in blacklist processing


peleion:
Is there any legitimate reason a server should be able to list itself with more than one DNS name for a single IP? The ongoing sybil attackers are vastly outnumbering the 80-90 legitimate servers by constantly flooding the peer list with multiple rapidly changing subdomains then changing them or the hostname sometimes several times daily and the blacklist.json cannot keep up. The recent push to immediately mark blacklisted hosts as bad and stop gossiping about them only fixes it _after_ the attacker has  been placed on the blacklist (and works well) but the blacklist appears to change about only once daily. UPDATE: blacklist is actually updated hourly, downloaded every 10 minutes and new PR will reapply blacklist to existing peers on every blacklist download.

The attackers show a preference for Digital Ocean droplets on 23 different subnets and change IPs and DNS  names sometimes hourly. I can verify personally that blacklisting these subnets (to reduce DoS abuse) returns the peerlist to nearly normal as I have done that but then get flagged as a reverse sybil for returning less than half the consensus known DNS peers as discussed with shsmith previously: https://github.com/spesmilo/electrum/issues/5091#issuecomment-464366843 UPDATE: this reverse sybil check is no longer active

If the ElectrumX blacklist processsing also cheked whether the requesting peer's IP was already included in the accepting peer's list and refuse to add it again if it were the ratio of attackers to legitimate servers would be vastly reduced and enhance safety for clients until the blacklist was updated within the next day. Also marking a peer as a sybil if say it returned more than 50% delta consensus increase in number of peers _based on DNS_ would be required if that's not done now with possibly a lower threshold (25%) for delta number of increase peer IPs as it would be unlikely for that many legitimate servers to come online at once. 

Note: this is not the same issue as #727 where duplicate DNS names are returned for the same IP - I'm suggesting refusing different DNS names for the same IP. 

Edits: grammar/clarity and updates

kyuupichan:
Patches would be nice; I'm not working on ElectrumX at the moment.  Lots of people have suggestions, very few want to write any code.

Note this whole problem is not an issue with ElectrumX; unfortunately Electrum chose the path they did to placate authors of wallet comparisons about privacy ... which backfired spectacularly.  Instead of giving an IP address to electrum.org, many users have now given up their coins instead.

peleion:
[...]

PR #758, opened 2019-03-13
peers: force retry blacklisted active peers after fetching blacklist

SomberNight:
implements https://github.com/kyuupichan/electrumx/issues/756#issuecomment-472550913
to address https://github.com/kyuupichan/electrumx/pull/753#issuecomment-471321214

So, after fetching the blacklist, force a rerun of `_should_drop_peer` and `_verify_peer` for each of our current peers that are in the newly fetched blacklist. This will ensure that blacklisted peers get marked bad immediately (instead of up to 24 hours later).


kyuupichan:
Thank you!

PR #761, opened 2019-03-15
Windows support for ElectrumX

sancoder:
Working on Windows support in electrumx


kyuupichan:
Re windows-support: that's neat.  Can you possibly bundle everything into one PR for that?

sancoder:
Will try to accomplish but no guarantees 

--
 Sincerely,
 Anton Shevchenko
 sancoder@gmail.com

PR #763, opened 2019-03-17
peers: case insensitive blacklist

bjd:
Fixes hosts bypassing the blacklist by varying capitalization in their hostnames.


kyuupichan:
Thanks

PR #764, opened 2019-03-18
Add support for Electra (ECA)

Jenova7:


Jenova7:
Build failed, also we use the `NIST5_hash` module which is not installed by default.
Could be that?
working well on my server

kyuupichan:
You need to add your python hash module to .travis.yml

PR #765, opened 2019-03-19
peers: small improvements against sybils

SomberNight:
1. IP based restrictions on internal peer list
    - forbid multiple peers sharing single IPv4
    - forbid multiple peers sharing IPv6 /64
    - limit number of `.onion` peers relative to total number of peers

With this change, the number of "good" peers went from ~450 to ~120 on a BTC mainnet server. 

2. also bucket IPv6 servers when replying to `server.peers.subscribe`
    - two servers per /56
3. small improvement to the blacklist code such that IP addresses can be banned even if "host" itself is not an IP

Re IPv6 subnets, I've read that residential end users often get a /64; but some ISPs give a /56. Some cheap VPSes only give a /128. In light of this, (if we ignore the cheap VPS case), the restrictions here are quite relaxed. We can make them stricter when required.

### DIFF
--- a/electrumx/lib/peer.py
+++ b/electrumx/lib/peer.py
@@ -25,7 +25,7 @@
 
 '''Representation of a peer server.'''
 
-from ipaddress import ip_address, IPv4Address, IPv6Address
+from ipaddress import ip_address, IPv4Address, IPv6Address, IPv4Network, IPv6Network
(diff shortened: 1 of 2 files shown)


kyuupichan:
This is getting complicated...

PR #785, opened 2019-04-05
WIP: naive per-IP session-resource limiting

SomberNight:
> Per IP address periodic resource usage limit.
  Very similar to :envvar:`BANDWIDTH_LIMIT`, with a notable difference
  being that this limit is based on the IP address of the session and
  is persisted (only in memory) across reconnects.
> 
> Requests made by sessions incur costs. This cost is a naive heuristic
  that tries to correlate with the CPU and DISK I/O usage of the request.
> 
> The time period is one day. The default resource limit is coin-specific.
  For BTC, it is 30000. Set to zero for "unlimited".

"WIP" mainly because I am still fiddling around with it a bit;
and I am not even sure whether this is the right approach...
Made the PR so that it can get feedback and others can test it too.

### DIFF
--- a/docs/environment.rst
+++ b/docs/environment.rst
@@ -274,6 +274,21 @@ raise them.
   Bandwidth usage is gradually reduced over time by "refunding" a
   proportional part of the limit every now and then.
 
+.. envvar:: RESOURCE_USAGE_LIMIT
+
+  Per IP address periodic resource usage limit.
+  Very similar to :envvar:`BANDWIDTH_LIMIT`, with a notable difference
+  being that this limit is based on the IP address of the session and
+  is persisted (only in memory) across reconnects.
+
+  Requests made by sessions incur costs. This cost is a naive heuristic
+  that tries to correlate with the CPU and DISK I/O usage of the request.
+
+  The time period is one day. The default resource limit is coin-specific.
+  For BTC, it is 30000. Set to zero for "unlimited".
+
+  See the :ref:`res_usage` RPC command.
+
 .. envvar:: SESSION_TIMEOUT
 
   An integer number of seconds defaulting to 600.  Sessions with no
--- a/docs/rpc-interface.rst
+++ b/docs/rpc-interface.rst
@@ -115,6 +115,27 @@ Low-priority sessions have their requests served after higher priority
 sessions.  ElectrumX will start delaying responses to a session if it
 becomes sufficiently deprioritized.
 
+res_usage
+---------
+
+Return the state of the internal per-session-IP resource limit tracker.
+As a form of DOS limit, requests made by sessions have costs.
+These costs are incurred by the IP address of the session.
+The result is sorted in decreasing order of costs::
+
+  $ electrumx_rpc res_usage
+  [
+    [
+        "41.200.217.53/32",
+        3624050.0391892944
+    ],
+    [
+        "187.39.240.113/32",
+        3599832.8948473367
+    ],
+    ...
+  ]
+
 groups
 ------
 
--- a/electrumx/lib/coins.py
+++ b/electrumx/lib/coins.py
@@ -86,6 +86,7 @@ class Coin(object):
     PEER_DEFAULT_PORTS = {'t': '50001', 's': '50002'}
     PEERS = []
     BLACKLIST_URL = None
+    RESOURCE_USAGE_LIMIT = 0
 
     @classmethod
     def lookup_coin_class(cls, name, net):
@@ -427,6 +428,7 @@ class BitcoinSegwit(BitcoinMixin, Coin):
     TX_COUNT_HEIGHT = 524213
     TX_PER_BLOCK = 1400
     BLACKLIST_URL = 'https://electrum.org/blacklist.json'
+    RESOURCE_USAGE_LIMIT = 30_000
     PEERS = [
         'btc.smsys.me s995',
         'E-X.not.fyi s t',
--- a/electrumx/server/env.py
+++ b/electrumx/server/env.py
@@ -72,6 +72,7 @@ class Env(EnvBase):
         self.max_sessions = self.sane_max_sessions()
         self.max_session_subs = self.integer('MAX_SESSION_SUBS', 50000)
         self.bandwidth_limit = self.integer('BANDWIDTH_LIMIT', 2000000)
+        self.res_usage_limit = self.integer('RESOURCE_USAGE_LIMIT', self.coin.RESOURCE_USAGE_LIMIT)
         self.session_timeout = self.integer('SESSION_TIMEOUT', 600)
         self.drop_client = self.custom("DROP_CLIENT", None, re.compile)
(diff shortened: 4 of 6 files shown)


whizz:
Awesome, thanks. I am running this now, will let you know how it works.

SomberNight:
I have several hundred IPs for which my server is sleeping over a minute before each request, I think after a certain threshold we should just disconnect from these. We could do something like this:

```patch
diff --git a/electrumx/server/session.py b/electrumx/server/session.py
index fb390e2..48959ee 100644
--- a/electrumx/server/session.py
+++ b/electrumx/server/session.py
@@ -735,6 +735,8 @@ class SessionBase(RPCSession):
         bucket = self._bucket_for_resource_usage()
         res_usage = self.session_mgr.res_usage_of_ip[bucket]
         sleep_time = res_usage / self.res_usage_limit - 1
+        if sleep_time > 30:
+            raise FinalRPCError(BAD_REQUEST, 'session is using too much resources')
         if sleep_time > 0:
             await sleep(sleep_time)
```

but this needs https://github.com/kyuupichan/aiorpcX/pull/15

peleion:
Seems not to work for me. Built using:
`pip3 install git+https://github.com/SomberNight/electrumx.git@res_usage_20190404`
and environment var manually set:
`RESOURCE_USAGE_LIMIT: '30000'`

electrumx_rpc res_usage after a few minutes:
`[
    [
        "181.196.52.141/32",
        112235.55579818286
    ],
    [
        "90.10.38.32/32",
        88802.32538499439
    ],
    [
        "85.241.21.171/32",
        77286.26852567006
    ],
    [
        "88.230.86.188/32",
        73226.47956282563
    ],
`

SomberNight:
You don't need to set it; it defaults to 30000.
In what way does it not work?

-------- Original Message --------

peleion:
Clarification: pull #786 aready merged lowers CPU tremendously but is not in this branch - you made it afterwards. Testing #785 then showed a significant jump in CPU so this masked the effect - in fact  performance is worse with only #785 than only #786 and no change in connected # of peers. Would be nice to be able to use both simultaneously. 

Also, I did not realize the res_usage showed IP resource use _after disconnection_ but this makes sense to me now - it has to be persistent in memory for 24 hours. However, abusive peers above the limit (30,000) are disconnected but then allowed to reconnect using multiple other ports (the MO of the botnet) but don't seem to be using additional resources. Weird. 

electrumx_rpc res_usage:

[
        "87.116.178.168/32",
        110324.9805268778
    ],

simultaneous electrumx_rpc sessions:

8451   SC6               3.3.1   1.4     0     0       0      13       1      13       2    06m13s  87.116.178.168:26522
8747   SC6               3.3.1   1.4     0     0       0      11       0      11       1    05m56s   87.116.178.168:8634

IPs above the limit appear to stay connected if there are subs - I assume this is intended. I set the environment variable for testing different values - it indeed defaults to 30,000 if unset.

gits7r:
I think it would be better to implement unique session_id identifiers for each session based on remote ip:port as well as some random cookie, so the resource usage costs could be endured by the session_id as opposite to IP address.

It's problematic for Tor servers for example, that are accessible via .onion hostnames. These servers see all their clients as connecting from 127.0.0.0/8 (localhost loopback) -- this is how Tor works. So, if you have 100 clients behind Tor and one abuser, e-x will disconnect all the honest clients as well.

I know that this source 127.0.0.1 from Tor is problematic, especially in cases  where you need to filter / drop / firewall stuff or defend against DoS but it's how anonymity works. I have very high usage on my servers via Tor, and most probably the attacker will also connect via Tor to attack my CPU resources, so I have no way to maintain my real honest clients that are not using resources abusively. That's why I was thinking of unique session_id, so we have 500 sessions from 127.0.0.1 but each has an unique ID.  I understand in practice this might not be as simple, because an attacker will connect, get an unique session_id, try to attack CPU resources, get disconnected, retry - this might be a good enough DoS as well? Still thinking about mitigation measures ...

SomberNight:
@gits7r what exactly is your goal with session IDs? There is already a concept of a session internally, and there is already bandwidth usage limiting for sessions, and btw these bandwidth limits (throttling) are not hit by the attacker as he makes a couple hundred expensive (cpu/disk) requests and then reconnects (before hitting the bw limit).

If you allow the attacker to choose his session ID (but I'm not sure exactly what you mean) then he can just keep doing the current attack and make sure to always get new session IDs.

And yes, what this PR is doing is insufficient for Tor, but I'm not sure how to handle that better. If you look at the change here, private IPs are exempt.
-------- Original Message --------

SomberNight:
Re .onion, the issue is that the requirements/ situation then become:

- anonymity for clients
- server serving expensive (disk/cpu) requests
- service being free

Choose two.
I don't see how you could have all three in a world with abusive clients. :(
-------- Original Message --------

gits7r:
@SomberNight you are right, there is no clear solution for the .onion situation described by me. Also, the clients that connect via Tor to clearnet electrumx servers (not .onion hostnames) will face the same problem as the servers will be seeing the IP address of the Tor exit node as the client IP address, and given there are hundreds of thousands of Tor users and some 950 exit relays (IP addresses), more and more clients get assigned the same exit node, and if just one client is abusive, others are prevented to connect too.

But as you said, this is the limitation -- we can't have all 3 in a world with abusive clients :( . Solution for .onion servers is to provide subscriber based service, that can be free as well if the operator chooses so and it can make the onion require authentication and provide the cookie to trusted clients.

EagleTM:
I'd vote for a default of 50000 rather than 30000 as the latter did block legitimate requests of restoring medium sized wallets.

I've achieved good results with closing sessions of clients above the resource limit rather right away rather than letting them sleep. The majority of attackers never reconnect, some reconnect violently but then gets dropped again within the timeout limit. It does not achieve tarpitting this way but frees up resources. 

 ```diff
 diff --git a/electrumx/server/session.py b/electrumx/server/session.py
 index fb390e2..48959ee 100644
 --- a/electrumx/server/session.py
 +++ b/electrumx/server/session.py
 @@ -735,6 +735,8 @@ class SessionBase(RPCSession):
          bucket = self._bucket_for_resource_usage()
          res_usage = self.session_mgr.res_usage_of_ip[bucket]
          sleep_time = res_usage / self.res_usage_limit - 1
+        if res_usage > self.res_usage_limit:
+            self.logger.info(f'close session over res usage for {ip_address(self.peer_address()[0])}')
+            raise RPCError(BAD_REQUEST, 'session is using too much resources')
+            await session.spawn(session.close())
          if sleep_time > 0:
              await sleep(sleep_time)
 ```

The logger spams a bit too much and might not be needed in production, just for debugging.

kyuupichan:
I'll close this; I think it's mostly superseded by recent commits.  If you think a new RPC is useful, or that session grouping or cost levels could be better, please submit a new PR. Thanks to SomberNight for motivating me to work on a similar idea.

PR #787, opened 2019-04-06
Change peer list refresh / stale timeout from 24 hours to 1 hour

peleion:
With the current DDoS attack the availability of servers changes often throughout the day (DDoS intensity, server load, server reboots) - well within 24 hours. Currently no server has an accurate reachable peer list since a newly unreachable server is still served as reachable for 24 hours due to the peer reachability / staleness test only being done once a day then retried multiple times before finally forgetting unreachable peers. 

Changing to hourly would help give clients a quicker syncing (and therefore better) experience since they wouldn't be trying to connect to servers that may be unreachable for up to 23 hours prior (due to server overload / DDoS, etc) with only a negligible increase in server load hourly.


kyuupichan:
I made it 3 hours; 1 is a bit too frequent

PR #793, opened 2019-04-11
Update CivX to EXOS for mainnet and testnet

turcol:


kyuupichan:
So this is just a rename?  What happened to the code removal?

turcol:
> So this is just a rename? What happened to the code removal?

This is just a rename. No code was removed

turcol:
Thanks !

PR #806, opened 2019-04-24
Bump default cost limits for AuxPoW coins

JeremyRand:
This is necessary because AuxPoW coins' headers are substantially larger, so the increased bandwidth usage during header download was hitting the default limits.

### DIFF
--- a/docs/environment.rst
+++ b/docs/environment.rst
@@ -256,9 +256,12 @@ raise them.
 .. envvar:: REQUEST_SLEEP
 .. envvar:: INITIAL_CONCURRENT
 
-  All values are integers. :envvar:`COST_SOFT_LIMIT` defaults to :const:`1,000`,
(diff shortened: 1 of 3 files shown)


kyuupichan:
Isn't it better to change the BW_UNIT_COST default instead?

JeremyRand:
> Isn't it better to change the BW_UNIT_COST default instead?

@kyuupichan Hmm, I think it depends on how the current weighting was determined.  I assumed that the weighting of the factors that contribute to the cost metric was based on how much each factor contributes to the load on the server.  If that's accurate, then I think this PR is correct.  If, on the other hand, the weighting was based on how much load should be contributed by a typically-behaving client, then I agree that changing the weighting is the correct fix.  Can you clarify on how the weighting was determined?

sidhujag:
Should also update other AuxPow coins, like Syscoin. I have been using these defaults 10x higher in environment variables but through the code is even better

kyuupichan:
Closing, no progress

PR #811, opened 2019-04-28
Added Bitcoin CZ

devcon-gh:
Test added


devcon-gh:
I'm sure my values are right, do I need some deserializer to convert hex to hash?

`>       assert coin.header_hash(
            block.header) == hex_str_to_hash(block_info['hash'])
E       AssertionError: assert b'\xd5\x80\xf...c\xb1\xb3\x89' == b'y\xb2\xab\x8...f\x00\x00\x00'
E         At index 0 diff: 213 != 121
E         Use -v to get the full diff`

PR #812, opened 2019-04-28
Add drop_peer RPC command

valesi:
Takes a hostname or IP as it appears in the peers list, and simply forgets it. That is, it does not mark it bad, so it can be added again through regular discovery or registration, permitting it verifies properly.


kyuupichan:
Why would you want to do this?

kyuupichan:
No feedback

PR #813, opened 2019-04-29
Added Cryptoescudo

VDamas:


kyuupichan:
If you want this in, please add tests and fix the failure.

PR #815, opened 2019-05-02
Add support for ODIN Coin

Manbearpixel:
Thanks for your work thus far on ElectrumX! I'd like to request support for adding ODIN Blockchain onto ElectrumX. We're launching mobile wallets and are currently using your implementation for these lightweight clients.


kyuupichan:
Thanks - can you fix the failure?  I use pycodestyle and it wants 2 line breaks between top-level class definitions.

Manbearpixel:
@kyuupichan Should be all fixed up now, thanks for your quick response!

kyuupichan:
The test you added is failing

Manbearpixel:
@kyuupichan Figured out the issue. Must have misunderstood what was needed for the test cases.

PR #827, opened 2019-05-18
Add SegWit Support for BitZeny

y-chan:
Add SegWit Support for BitZeny and change header hash yescrypt to yespower


kyuupichan:
No sys exit please.

y-chan:
added: fix import error and fee

PR #830, opened 2019-05-27
Add support for estimation mode in estimatesmartfee

shyrwall:
Suggestion to add support for estimation modes in the estimate fee cmd. This would allow for example the Electrum client to implement support for choosing ECONOMICAL or CONSERVATIVE (default) mode. Conservative is often flawed and results in a higher fee suggestion than necessary.

Personally I want ECONOMICAL to be the default but that would be an issue on the Electrum side.

If accepted please review code. Maybe there is a better way of doing this.

### DIFF
--- a/electrumx/server/daemon.py
+++ b/electrumx/server/daemon.py
@@ -228,12 +228,12 @@ class Daemon(object):
         '''Update our record of the daemon's mempool hashes.'''
         return await self._send_single('getrawmempool')
 
-    async def estimatefee(self, block_count):
+    async def estimatefee(self, block_count, estimate_mode):
         '''Return the fee estimate for the block count.  Units are whole
         currency units per KB, e.g. 0.00000995, or -1 if no estimate
         is available.
         '''
-        args = (block_count, )
+        args = (block_count, estimate_mode)
         if await self._is_rpc_available('estimatesmartfee'):
             estimate = await self._send_single('estimatesmartfee', args)
             return estimate.get('feerate', -1)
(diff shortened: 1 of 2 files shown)


kyuupichan:
When was this argument introduced to bitcoind?  Was it always there?  We have to consider bitcoind implementations other than core.

Also please fix the test failures

shyrwall:
Did not think about other implementations. It did exist in ABC for example but was removed. But it has now been removed. Guess I need to rethink this and make it Core specific somehow.

SomberNight:
It would probably be easiest to send no argument for bitcoind if no argument was specified to e-x; i.e. to keep current behaviour unless the argument was actually specified.

kyuupichan:
When I was asking about the argument, I meant in core - was it always present with estimatesmartfee?  Note the existence of the RPC call is already checked, so implementations that don't have it are not an issue.  If the argument was there from the beginning then all the others (e.g. LTC) that copy core would have it too, and so would older core daemons.

shyrwall:
https://github.com/bitcoin/bitcoin/commit/104f5f21dc2413254dc9ae3651e10a5c0730c951

Seems it was merged in July 2017.

SomberNight:
I think this arg is only available for `estimatesmartfee`, and I think that method have always had it.

kyuupichan:
Does anyone intend to make this production-worthy for all coins?  If not I'll close it.  Note there are daemon tests, they should be updated too.

PR #840, opened 2019-06-16
NIX Sigma implementation

phamels:
Custom deserializer for NIX with support for Sigma


kyuupichan:
Testcase?

phamels:
Tests are already in place as NIX was already added to Electrumx but this is a modification to the previous implementation.

kyuupichan:
You've changed to a new serializer.  Any old tests can't be testing that otherwise they would presumably have failed?

phamels:
Only test implemented is a block test which is still valid for the block it was testing

kyuupichan:
What I'm saying is you've changed code.  You have no tests the change is good.

PR #849, opened 2019-06-24
Added Ritocoin

traysi:
This PR adds support for Ritocoin and the X21S hashing algorithm.

PR #854, opened 2019-06-27
Readd Ritocoin support

streetcrypto7:
As per the build error is solved https://github.com/kyuupichan/electrumx/pull/849#issuecomment-506315873


streetcrypto7:
Test passed https://travis-ci.org/kyuupichan/electrumx/builds/551415562

PR #867, opened 2019-07-11
BITG masternode support

konez2k:
Updated also the BitcoinGreen rebrand to BitGreen


kyuupichan:
The code copy-paste isn't acceptable; use inheritance.  Also remove your .coverage file.

Hunghus:
hello @konez2k We need fork pivx. Can you help us? We will pay you. We need urgently. Save us

konez2k:
@kyuupichan I'll try to base it from Dash, excluding the masternode list the other commands should be compatible

konez2k:
@Hunghus this is not the channel, discord konez2k#6060

konez2k:
sorry @kyuupichan, I'll leave this open for 2 or 3 weeks, because BITG is changing also the masternode list.

electrumx/server/daemon.py
```diff
@@ -496,3 +496,14 @@ def strip_mtp_data(self, raw_block):
         blocks = await self._send_vector('getblock', params_iterable)
         # Convert hex string to bytes
         return [hex_to_bytes(self.strip_mtp_data(block)) for block in blocks]
+
+
+class BitGreenDaemon(Daemon):
```
kyuupichan: Please rename this DashDaemonBase, and have DashDaemon and SmartCashDaemon inherit it.

electrumx/server/session.py
```diff
@@ -1676,3 +1676,137 @@ def truncate_auxpow(self, headers_full_hex, start_height):
             height += 1
 
         return headers.hex()
+
+
```
kyuupichan: Similarly here - this is common to all dash derivatives so it should be DashElectrumXBase and have DashElectrumX inherit it.  Then there's no copy-paste.

PR #873, opened 2019-07-24
Share more code for name-prefix handling

domob1812:
Namecoin and Emercoin both have special treatment for name prefixes on scripts.  Through the NameMixin, they were already sharing some code for that.  There was, however, still a lot of duplicated logic.

This refactors the code so that more logic is shared in the common `NameMixin`.  In particular, there is now a general function for matching a potential name prefix against the allowed structures of name operations.  Thus, the coin-specific code only has to define the valid structures of name operations in the coin and handle coin-specific extra work (like building the normalised script that is used for indexing names in Namecoin).

### DIFF
--- a/electrumx/lib/coins.py
+++ b/electrumx/lib/coins.py
@@ -40,7 +40,8 @@ from functools import partial
 import electrumx.lib.util as util
 from electrumx.lib.hash import Base58, hash160, double_sha256, hash_to_hex_str
 from electrumx.lib.hash import HASHX_LEN, hex_str_to_hash
-from electrumx.lib.script import ScriptPubKey, OpCodes
+from electrumx.lib.script import (_match_ops, Script, ScriptError,
+                                  ScriptPubKey, OpCodes)
 import electrumx.lib.tx as lib_tx
 import electrumx.lib.tx_dash as lib_tx_dash
 import electrumx.server.block_processor as block_proc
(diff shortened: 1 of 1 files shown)


domob1812:
I plan to add support for [Xaya](https://github.com/xaya/xaya) in the near future, which will have name handling similar to Namecoin except that it supports slightly different name operations.  Thus with this change, implementing Xaya will be easier and won't create yet another copy of the duplicated code.

domob1812:
Rebased to solve merge conflicts with #875.

domob1812:
@kyuupichan, can you please take a look at this and let me know if it can be merged?  I have more changes in store which will add full support for [Xaya](https://xaya.io/).  I believe that it is easier to review smaller, self-contained PRs (and this is a first one).  If you would rather have all changes at once, let me know and I can create one big PR instead.  Thanks!

electrumx/lib/coins.py
```diff
@@ -369,6 +372,58 @@ def find_end_position_of_name(script, length):
...
+        fields (e.g. the name) returned.
+
+        possible_ops must be an array of arrays, defining the structures
+        of name prefixes to look out for.  Each array can consist of
+        actual opcodes, -1 for ignored data placeholders and strings for
+        named placeholders.  Whenever a data push matches a named placeholder,
+        the corresponding value is put into a dictionary the placeholder name
+        as key, and the dictionary of matches is returned."""
+
+        from electrumx.lib.script import _match_ops, Script, ScriptError
```
kyuupichan: This might be better added to the module-level import of the same?

domob1812: I just kept this how it was in the code I refactored, but I agree that it makes sense to move that to the module level.  Done.

domob1812:
> It would be nice if some simple tests that interpret_name_prefix is working as expected for its various use cases.
I can do that, but it seemed to me the main way of testing things like this in the existing code was through the tx tests.  These cover all cases of name operations, so they also ensure that the code here works as it should for all relevant situations.

But if you want me to add an explicit test instead, just let me know and I'll do it.

PR #880, opened 2019-08-10
Add clarifications for how to allow electrumx to bind to ports < 1024

gits7r:
On most systems a python3 executable will not be able to bind to ports lower than 1024 without some additional settings, better to explain this in documentation. Otherwise, ElectrumX will look like it's running as configured but without actually listening on the desired ports, if they are lower than 1024.


kyuupichan:
Thanks, I committed something a little briefer here:

https://github.com/kyuupichan/electrumx/commit/5fe8b2f9d25ee4125e90b1a35eaa3e7c4f4134b0

PR #889, opened 2019-08-29
Add Xaya as new coin with basic chain handling

domob1812:
[Xaya](https://xaya.io/) is based on Namecoin, but differs in two main aspects:  First, the chain structure is different, with a [custom model for PoW mining](https://github.com/xaya/xaya/blob/master/doc/xaya/mining.md) that is based on auxpow but not exactly the same as in Namecoin.  Second, name operations in Xaya are based on Namecoin, but differ in details (names do not expire, registration of names works differently).

In this PR, we add Xaya as a coin and the things relevant for its basic chain structure (mainly the new PoW format).  Enabling name support in Xaya will come as a second PR afterwards.


domob1812:
Since the chain and name aspects of Xaya are pretty orthogonal, I think it is easier to review both parts separately in smaller PRs than having one large PR with everything.  @kyuupichan, let me know if you'd rather have only one big PR instead.

kyuupichan:
This is fine - but please placate pycodestyle; it's failing with your change (see travis above)

domob1812:
Fixed the style error.  The remaining travis error seems to be unrelated to this PR as far as I can tell.

PR #895, opened 2019-09-20
Ravencoin: Switch algo to x16rv2 after activation date

standard-error:
Ravencoin mainnet is forking to x16rv2 on Oct 1st (already active on testnet)
https://twitter.com/ravencoin/status/1167469138178125825?lang=en

Also updates the txcounts and removes dead public servers (so, err, all of them)


kyuupichan:
Test failed with unknown import

standard-error:
x16rv2_hash breaking in travis environment although works ok here locally. Not my module so I'll ask the author to look at it.

PR #899, opened 2019-09-25
Fix Emercoin nvs values parse

yakimka:
For now, for scripts in which the value takes more than one placeholder, the address_script is parsed incorrectly.
Emercoin has such script structure:
```
(name_new | name_update) << OP_DROP << name << days << OP_2DROP << val1 << val2 << .. << valn << OP_DROP2 << OP_DROP2 << ..<< (OP_DROP2 | OP_DROP) << paytoscripthash
```
Therefore, a Emercoin can contain an unlimited number of placeholders with a value.

To fix this, I changed the method `interpret_name_prefix` of `NameMixin` and added a new opcode `-2` which means "Data Push Multiple". I also added tests for `NameMixin`.


kyuupichan:
@domob1812 does this look good to you?

domob1812:
I've not looked at the tx structure for Emercoin in detail, but as long as it doesn't break the existing tests for Namecoin or Xaya, it is fine for us.  Perhaps there could be another Emercoin example transaction added for the tests (which exhibits this structure), instead of or in addition to the unit test in code?

yakimka:
Emercoin transaction which exhibits this structure is already in the repository.
Added tests are needed to ensure that the script is split correctly because the `test_transactions.py` does not guarantee this.

domob1812:
@yakimka That makes sense.  For Namecoin/Xaya, the correctness is tested by comparing `hashX_from_script` to the addresses reported in the transaction data in `test_transactions.py`.  Perhaps you can do something like that as well?

But that is just a random thought - I have certainly no objection to your extended coins test.

yakimka:
@domob1812 yes you are right. I added another test transaction for a case that was not covered before. But still, I think it makes sense to leave extended coin tests because `test_transactions.py` doesn't cover all cases.

PR #902, opened 2019-10-04
adds Dockerfile

breign:
For easier deployment on different machines and environments, I'm proposing a Dockerfile.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -1,0 +1,35 @@
+# example of Dockerfile that builds release of electrumx-1.13.0
+# ENV variables can be overrided on the `docker run` command
+
+FROM ubuntu:18.04
+
+WORKDIR /
+ADD https://github.com/kyuupichan/electrumx/archive/1.13.0.tar.gz /
+RUN tar zxvf *.tar.gz
+
+RUN apt-get update && \
+        apt-get -y install python3.7 python3-pip librocksdb-dev libsnappy-dev libbz2-dev libz-dev liblz4-dev && \
+        pip3 install aiohttp pylru python-rocksdb
+
+RUN cd /electrumx* && python3 setup.py install
+
+ENV SERVICES="tcp://:50001"
+ENV COIN=BitcoinSV
+ENV DB_DIRECTORY=/db
+ENV DAEMON_URL="http://username:password@hostname:port/"
+ENV ALLOW_ROOT=true
+ENV DB_ENGINE=rocksdb
+ENV MAX_SEND=10000000
+ENV BANDWIDTH_UNIT_COST=50000
+ENV CACHE_MB=2000
+
+VOLUME /db
+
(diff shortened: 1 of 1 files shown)


contrib/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# example of Dockerfile that builds release of electrumx-1.13.0
+# ENV variables can be overrided on the `docker run` command
+
+FROM ubuntu:18.04
+
+WORKDIR /
+ADD https://github.com/kyuupichan/electrumx/archive/1.13.0.tar.gz /
```
streetcrypto7: 🤔 Could we change this to something like git clone https://github.com/kyuupichan/electrumx?

contrib/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN tar zxvf *.tar.gz
+
+RUN apt-get update && \
+        apt-get -y install python3.7 python3-pip librocksdb-dev libsnappy-dev libbz2-dev libz-dev liblz4-dev && \
+        pip3 install aiohttp pylru python-rocksdb
+
+RUN cd /electrumx* && python3 setup.py install
+
+ENV SERVICES="tcp://:50001"
+ENV COIN=BitcoinSV
```
streetcrypto7: BitcoinSV -> BitcoinSegwit please

contrib/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ENV DB_DIRECTORY=/db
+ENV DAEMON_URL="http://username:password@hostname:port/"
```
streetcrypto7: lmao where is the bitcoind?

contrib/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ENV DB_ENGINE=rocksdb
+ENV MAX_SEND=10000000
+ENV BANDWIDTH_UNIT_COST=50000
+ENV CACHE_MB=2000
+
+VOLUME /db
+
+RUN mkdir -p "$DB_DIRECTORY" && ulimit -n 1048576
+
+CMD ["/usr/bin/python3", "/usr/local/bin/electrumx_server"]
```
streetcrypto7: Hmm shouldn't be /usr/bin/python3.7 instead of /usr/bin/python3?

contrib/Dockerfile
breign: ENV can also be set at `docker run` command to override anything to their needs.

contrib/Dockerfile
breign: Ubuntu distro has a symlink to their python 3.* installation.

contrib/Dockerfile
breign: This is all configurable at run.

contrib/Dockerfile
breign: Please do. I was just pasting a Dockerfile that works for me.

PR #910, opened 2019-10-19
Adding Veil

MatWaller:
Adding support for [Veil](http://www.veil-project.com)


MatWaller:
Found a issue -- resolving and i will resubmit.

PR #920, opened 2019-11-06
Add Vitae Support

akshaynexus:


akshaynexus:
image
Tested and working

kyuupichan:
Please remove electrum_header; it is dead code.  Also add block tests like other coins

akshaynexus:
Ok,but how would i parse the header extras then?

kyuupichan:
I don't understand the question.  That function is not used; it was for returning stuff to the client, but the RPC call no longer exists.

akshaynexus:
Oh ok,thanks,will do the necessary changes today

Issue #934, opened 2019-12-16
ModuleNotFoundError: No module named 'aiorpcx'


pedr0-fr:
Running eletrumX for the first time. I followed the [guide](https://github.com/kyuupichan/electrumx/blob/master/docs/HOWTO.rst) and everything went smoothly: installed pre-requisites, cloned it, ran setup.py, set up the configuration files and started it with `systemctl start electrumx`. When I checked the log via `journalctl -u electrumx` I get:
```
Dec 16 01:46:11 lenovoW541 systemd[1]: Started Electrumx.
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]: Traceback (most recent call last):
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]:   File "/media/pedro/Portable_2TB/Bitcoin/electrumx/electrumx_server", line 16, in <module>
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]:     from electrumx import Controller, Env
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]:   File "/media/pedro/Portable_2TB/Bitcoin/electrumx/electrumx/__init__.py", line 4, in <module>
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]:     from electrumx.server.controller import Controller
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]:   File "/media/pedro/Portable_2TB/Bitcoin/electrumx/electrumx/server/controller.py", line 10, in <module>
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]:     from aiorpcx import _version as aiorpcx_version, TaskGroup
Dec 16 01:46:11 lenovoW541 electrumx_server[28383]: ModuleNotFoundError: No module named 'aiorpcx'
Dec 16 01:46:12 lenovoW541 systemd[1]: electrumx.service: Main process exited, code=exited, status=1/FAILURE
Dec 16 01:46:12 lenovoW541 systemd[1]: electrumx.service: Failed with result 'exit-code'.
```

I have checked that this package is installed (which was confirmed durin the setup) don't know how I can possibly get this error. I have aiorpcx version 0.18.4 installed on python 3.7.3 on Ubuntu 18.04.3.

SomberNight:
Make sure it's installed for the correct python version.

> python 3.7.3

Ubuntu 18.04  by default comes with python 3.6.x, so that means you have two different python interpreters at least.

[...]

PR #935, opened 2019-12-16
Note about anaconda

pedr0-fr:
As mentioned in https://github.com/kyuupichan/electrumx/issues/934


kyuupichan:
Hmm, not convinced it's the issue, but OK.

PR #974, opened 2020-03-13
Errors in the prefetcher can be hidden by subsequent errors in flushing.

rt121212121:


electrumx/server/block_processor.py
```diff
@@ -675,6 +675,10 @@ def spend_utxo(self, tx_hash, tx_idx):
             async with TaskGroup() as group:
                 await group.spawn(self.prefetcher.main_loop(self.height))
                 await group.spawn(self._process_prefetched_blocks())
+        except Exception:
+            # If there is an error in the prefetcher chances are the flush can error too, hiding
+            # this error.
+            self.logger.exception('error in prefetcher')
```
SomberNight: shouldn't you re-raise?

rt121212121:
Ah, it was late, very good point.

rt121212121:
I think this possibly needs to filter out logging for `CancelledError`.

kyuupichan:
In git HEAD I'm only flushing for CancelledError as others may be a consequence, or cause, of corrupted in-memory data which we don't want to flush and which tends to trigger the assertions in the flush code

PR #990, opened 2020-03-30
Add Syscoin

sidhujag:


sidhujag:
Note: chunk size was removed in all coins using it except syscoin because it was not actually used anywhere. Bitcoin's default was being used for all coins before this change here:

https://github.com/kyuupichan/electrumx/pull/990/files#diff-378cc43eb54e94fd61ed8589bedd056cR832

So now chunk being defined would be used, and in Syscoin diff retarget is 360 so that is the chunk size we want to use, which matches on the electrum client. (if we don't use proper chunk sizing it seems its reflective of SPV security but also checkpoints have issues because they assume chunks and retarget periods are consistent with each other)

sidhujag:
since Syscoin has AuxPow, we should also look at https://github.com/kyuupichan/electrumx/pull/806 and add in those defaults for sys similar to nmc

sidhujag:
@kyuupichan have you taken a look? Is there feedback for me to fix something?

kyuupichan:
It looks fine; I usually only review PRs in batches at irregular intervals, typically pre-release.  However your PR has conflicts; can you resolve them please.

sidhujag:
Why was this reverted?

kyuupichan:
The test was failing

sidhujag:
Ok let me check it out

PR #992, opened 2020-03-31
Add multiple coins

onnateldome:


electrumx/lib/coins.py
```diff
@@ -3793,3 +3793,787 @@ class Quebecoin(AuxPowMixin, Coin):
...
+            'version': version,
+            'prev_block_hash': hash_to_hex_str(header[4:36]),
+            'merkle_root': hash_to_hex_str(header[36:68]),
+            'timestamp': timestamp,
+            'bits': bits,
+            'nonce': nonce,
+        }
+
+
+class Lbry(Coin):
```
cipig, line 3931: this class is defined twice, see previous one

PR #994, opened 2020-04-02
[Session] Bump Proto Version and add Json Headers (WIP)

ahmedbodi:


ahmedbodi:
you lost me at the function of the protocol version part, mind explaining?

JeremyRand:
Lots of whitespace changes here that probably shouldn't be there.

JeremyRand:
It may be desirable to check the `STATIC_BLOCK_HEADERS` attribute of the coin, and split the headers differently based on that.  If the headers are static, you'll want to split them based on `BASIC_HEADER_SIZE`.  If they're dynamic, you'll want to split them based on `dynamic_header_offset`.  I suspect trying to do either with the wrong header type will break things.

JeremyRand:
Also it would probably be good to have a function available that always returns the array form (regardless of protocol version).  This makes it easier to implement the `AuxPoWElectrumX` version cleanly, i.e. `AuxPoWElectrumX` should call the superclass function that always returns the array form, do whatever truncation is needed to the individual headers, and then (if the concatenated form is desired in the final result) concatenate the truncated individual headers.  This would result in much cleaner (and smaller) AuxPoW code.

JeremyRand:
(I'll defer to Neil if he disagrees with any of the above review.)

electrumx/server/session.py
```diff
@@ -1170,6 +1170,8 @@ async def block_headers(self, start_height, count, cp_height=0):
         start_height and count must be non-negative integers.  At most
         MAX_CHUNK_SIZE headers will be returned.
         '''
+        if self.protocol_tuple >= (1, 5, 0):
+            return self.block_headers_array(start_height, count, cp_height)
```
JeremyRand, line 1174: Should this `await`?

electrumx/server/session.py
```diff
@@ -1186,6 +1188,39 @@ async def block_headers(self, start_height, count, cp_height=0):
...
+            last_height = start_height + count - 1
+            result.update(await self._merkle_proof(cp_height, last_height))
+
+        height = 0
+        while cursor < len(headers):
+            header_size = self.db.dynamic_header_len(height)
+            header = headers[cursor:cursor + header_size]
+            result['headers'].append(header)
+            cursor += header_size
+            height += 1
```
JeremyRand, line 1219: `dynamic_header_len` will yield an `assert` error for chains with static header length.  It's also unnecessarily expensive to use it for this, because its implementation checks both the start and end positions of each header, so this code `cursor + header_size` is actually doing the redundant calculation `cursor + self.dynamic_header_offset(height + 1) - cursor`.

I'd suggest replacing `header_size` with `next_cursor`, setting it to `self.db.header_offset` (which works for both static and dynamic lengths), and changing the slicing/incrementing accordingly.

JeremyRand:
Posted some inline comments.  Looks a lot better now.  Also note that you'll need to refactor `AuxPoWElectrumX.block_headers`, otherwise things will break for AuxPoW coins.  My recommendation is to make `AuxPoWElectrumX.block_headers` get the result of `super().block_headers_array`, pass it to `truncate_auxpow` if protocol is at least `(1, 4, 1)` and `cp_height != 0` (`truncate_auxpow` can be made a lot simpler since it'll just be looping through the array and truncating each header), and then concatenate the array if protocol is less than `(1, 5, 0)`.

kyuupichan:
My understanding is this is not production-ready; I intend to do a final release

ahmedbodi:
sure, do you have a rough idea when the next release is? been held up recently due to ramadhan and lockdown so been tough finding the time to work on stuff

PR #1001, opened 2020-05-11
 allow passing estimate_mode to estimatefee

cipig:


SomberNight:
I haven't tested but this LGTM.
Note that these type of changes typically involve bumping the protocol version and a mention in the docs.

cipig:
it is an optional parameter, so still works if you don't pass it:
```
(echo '{ "id": 1, "method": "blockchain.estimatefee", "params": ["1"] }'; sleep 0.5) | ncat 95.216.36.138 10000
{"jsonrpc": "2.0", "result": 0.00059318, "id": 1}

(echo '{ "id": 1, "method": "blockchain.estimatefee", "params": ["1","ECONOMICAL"] }'; sleep 0.5) | ncat 95.216.36.138 10000
{"jsonrpc": "2.0", "result": 0.00022116, "id": 1}
```

kyuupichan:
Requires doc update

PR #1026, opened 2020-07-10
use requirements.txt in place of a list in setup.py 'install_requires'

AustEcon:
- makes shell scripting easier for electrumsv-sdk to workaround issues on windows with installing plyvel


AustEcon:
Thanks 😉

Obmorok070 (commented):
Ghjjjj

PR #1030, opened 2020-07-27
Fix typo in LTORBlockProcessor::backup_txs

reedriley:
I'm not entirely sure what triggers this code; but I caught it failing.  Dumping the object confirmed it was an instance of `TxOutput`, and that the line in question would always fail.  Fixing the bad attribute name seemed to fix things.


kyuupichan:
This code has been removed from the codebase.

PR #1056, opened 2021-07-20
Features/tsc merkle proofs

AustEcon:


electrumx/server/session.py
```diff
@@ -643,7 +643,7 @@ def extra_cost(self, session):
             return 0
         return sum((group.cost() - session.cost) * group.weight for group in groups)
 
-    async def _merkle_branch(self, height, tx_hashes, tx_pos):
+    async def _merkle_branch(self, height, tx_hashes, tx_pos, return_root=False):
```
kyuupichan, line 646: I don't like functions returning differently-sized tuples based on how they're called.  Please return the root always now, and update the former callers to expect and ignore it.

kyuupichan (changes requested):
The new method also needs documenting in the docs/ directory.

Is adding tests too difficult?  If it is, it's not a problem.

electrumx/server/session.py
AustEcon: Okay no problem. No adding tests is okay - I still have local changes I haven't pushed. The main niggle is with adding asterixes in place of "duplicate hashes" in the branch but I will see what you think after I push the next commit(s)...

Choicezee01 (approved):

Choicezee01 (changes requested):
34xp4vRoCGJym3xR7yCVPFHoCNxv4Twseo

